PR #626, opened 2026-07-16
Implement Backtest menu: hardcoded "CAC40 Bougie de 9h" strategy

RNiveau:
## Summary

Implements the "Backtest" menu spec'd and merged in #624: a new menu that runs one hardcoded strategy — **CAC40 Bougie de 9h** (reuses the existing `Strategy.B9H` enum) — against historical FRA40.I candles. No generic backtest engine; this strategy's rules are fixed in code per FR-002.

Strategy recap:
- Reference range: the 9:00–10:00 (Paris local, DST-aware) H1 candle's high/low on FRA40.I.
- After 10:00, evaluate 5-minute candles: if price breaks below the H1 low and a later candle closes back above it, enter long at that close.
- Exit on whichever comes first: -50pts stop-loss, a stop moved to break-even once +20pts in profit, +[H1 high - 10pts] take-profit, or end of day. Multiple sequential (non-overlapping) trades per day are allowed.
- UI supports a single-day run (full trade detail) or a date range (aggregate summary: number of days, number of trades, winning/losing/BE positions, average win, average loss, final result), computed on demand, not persisted.

## What's included

**Backend**
- `model/backtest.py` — `BacktestDefinition`, `Trade`, `DayResult`, `DayResultSummary`, `BacktestSummary`, `BacktestRunResult`
- `model/enum.py` / `model/workflow.py` — new `ExitReason`, `DayStatus` enums, `UnitTime.M5`
- `services/candles_service.py` — `get_candles_in_window`, fetching candles over an explicit historical UTC window at a given horizon
- `api/services/backtest_service.py` — DST-aware Paris time helpers (`zoneinfo`), `evaluate_day` (the full rule engine), `run_range` (aggregation)
- `api/routers/backtest.py` — `GET /api/backtest/definitions`, `/day`, `/run`, with FR-016 date validation

**Frontend**
- `pages/Backtest.tsx` — single-day / date-range mode toggle, client-side validation
- `components/BacktestDayDetail.tsx` — reusable detail view (H1 range, trades, collapsible 5-minute candle sequence)
- Sidebar entry, route, `backtestService` in `api.ts`

**Tests**
- 41 new backend tests (`tests/api/services/test_backtest_service.py`, `tests/api/routers/test_backtest.py`, additions to `tests/services/test_candles_service.py`)
- 100% line coverage on every new backend file
- Explicit coverage of the two documented same-candle break-even edge cases (arm-vs-exit priority, arm-and-breach round-trip)

## Verification

- Full backend suite: 305 passed, 10 skipped (14 pre-existing AWS-context-dependent failures in `test_watchlist.py`/`test_workflow.py`, reproduced identically against a clean `main` checkout — unrelated to this change)
- `black`, `isort`, `flake8`, `mypy` clean on every new/changed file (two pre-existing `flake8` E501s and one pre-existing `mypy` aioboto3 stub warning left untouched — confirmed identical on `main`)
- Frontend `build`/`lint`: same pre-existing errors as `main` (6 `tsc` errors, 15 `eslint` problems), none touching the new Backtest files — diffed against `main` to confirm
- Manually verified in a real browser (Playwright): single-day run, range run, client-side date validation, and the range-to-day-detail drill-down all render and behave correctly

**Known gap**: `specs/021-backtest-menu-hardcoded/quickstart.md`'s SC-002 section (hand-verifying ≥6 *real* historical FRA40.I trading days against the actual chart) could not be performed in this sandbox — there's no real Saxo API credential available here, only `MockSaxoClient`. That real-data validation is still outstanding before this ships to production use.

## Test plan

- [x] Backend unit/router tests pass, 100% coverage on new code
- [x] Frontend build/lint show no new errors vs `main`
- [x] Manual browser verification of single-day, range, validation, and drill-down flows
- [ ] Hand-verify ≥6 real historical FRA40.I days against the actual chart (requires real Saxo credentials — not done here)

---
_Generated by [Claude Code](https://claude.ai/code/session_012S2PLVTLEPLCB15S2FV4j6)_


api/services/backtest_service.py
```diff
@@ -0,0 +1,362 @@
...
+            stop_level = position.stop_level
+            if candle.lower <= stop_level:
+                exit_price = (
+                    candle.open if candle.open <= stop_level else stop_level
+                )
+                reason = (
+                    ExitReason.BREAK_EVEN
+                    if position.be_armed
+                    else ExitReason.STOP_LOSS
+                )
```
RNiveau, line 392: **Break-even exit can produce non-zero points — the "BE is always exactly 0" invariant is not actually guaranteed.**

When `be_armed` is true, `stop_level == entry_price`. If a candle opens below entry (`candle.open <= stop_level`), the gap branch sets `exit_price = candle.open`, so `points = open - entry < 0`, yet `exit_reason` is still `BREAK_EVEN`. In `_build_summary` this trade is counted as BE regardless of points, so its negative points flow into `final_result` but not into `average_loss` — and it violates FR-013's "Number of BE … each with a points result of exactly 0".

This is really a spec-internal contradiction: FR-010 says gap-through-break-even records the exit at the candle open, while FR-008a / the Assumptions say a BE close is *always* 0 points. The code faithfully follows FR-010, which is defensible, but the two can't both hold. Worth deciding explicitly which wins:
- If FR-010 wins (current behavior): update FR-008a/Assumptions/FR-013 wording, since BE points can be negative.
- If "BE = 0" wins: clamp the BE exit to `entry_price` (no gap fill for the break-even case).

Reachable in a continuous session whenever a 5-min candle opens a point or two below entry while armed — small magnitude, but it does break the stated invariant.

---

RNiveau (commented):
## Deep review — Backtest menu ("CAC40 Bougie de 9h")

Read the full backend (`backtest_service.py`, router, models, `candles_service.get_candles_in_window`), the frontend page/component/service, and all three test files, cross-checked against `specs/021-backtest-menu-hardcoded/spec.md`.

**Overall: this is a strong, high-fidelity implementation.** The rule engine tracks the spec closely, including the genuinely tricky bits:

- **DST handling is correct.** `paris_reference_window_utc` converts 9:00–10:00 *Paris local* to naive-UTC bounds via `zoneinfo`, and Saxo's candle `Time` is parsed to naive UTC (`strptime` drops the `Z`), so the `start_utc <= date < end_utc` window filter selects exactly the right H1/5-min candles in both CET and CEST. 👍
- **Same-candle ordering (FR-009 / FR-008a) is right.** Exit checks run stop → take-profit → BE-arm as mutually-exclusive branches, and `be_armed` is only read via `position.stop_level` at the *start* of each candle, so a newly-armed break-even only bites from the next candle — the "arm-and-breach round-trip" and "arm-vs-exit priority" edge cases both behave as specified, and both are explicitly tested.
- **Gap-fill, multi-trade re-entry, NO_DATA vs NO_TRADE, and the range aggregation** all match the spec. Tests are meaningful (real numeric assertions, not mock-shaped), and the two documented BE edge cases are covered directly.

### Findings (all minor — none blocking)

1. **Break-even exit ≠ always 0 points** (inline comment on `backtest_service.py`). Gap-through-BE fills at the candle open, which can be negative while still labelled `BREAK_EVEN`. This is a spec-internal contradiction (FR-010 vs FR-008a/FR-013) worth resolving one way or the other.

2. **`run_range` fetches every calendar day sequentially, weekends included.** Each day = 2 `get_candles_in_window` calls → 2 `get_historical_data` HTTP fetches (the `get_asset` UIC lookup is `@lru_cache`d, so that part is fine, but historical-data results aren't shared across days). Since the spec deliberately imposes no max range, a multi-month/year run is a long sequence of blocking calls, with weekend/holiday days spending a fetch just to resolve to NO_DATA. Functionally correct, but a large range could be slow and hammer Saxo. Consider at least skipping `weekday() >= 5`, and/or noting the practical range limit in the UI.

3. **"Today" (an incomplete session) isn't blocked.** Validation only rejects *future* dates (`is_future_paris_date`), so a same-day run mid-session is allowed. Saxo won't return the in-progress H1/5-min candle and the day isn't closed, so results could be partial/unstable — which contradicts the spec's "operates only on already-closed historical days" assumption. Either reject `date >= today` or document that today yields partial results.

4. **Nit:** `Backtest.tsx` computes `today` from `new Date().toISOString()` (UTC) for its client-side future check, while the backend uses Paris-local — a harmless 1-day boundary discrepancy since the backend re-validates, but worth aligning.

The PR's own "known gap" is the important one operationally: the strategy has never been run against **real** FRA40.I data (only `MockSaxoClient`), so SC-002 (≥6 hand-verified historical days across all outcome types) is still outstanding. Given how much of this strategy's correctness lives in candle-boundary/DST/gap details that mocks can't surface, I'd treat that real-data pass as a hard gate before this drives any live decision — everything above is cosmetic next to it.

---

api/services/backtest_service.py
RNiveau: Confirmed and resolved in favor of FR-010 (keep current code behavior — no code change needed).

Updated `spec.md` (FR-008a, FR-010, FR-013, Assumptions) and `data-model.md`'s validation rule to state explicitly: gap-fill applies uniformly to all three exit types including break-even, and "break-even" is a classification of *which mechanism* closed the trade, not a guarantee the outcome was exactly flat. Added two regression tests locking this in — a gapped BE exit that keeps `exit_reason == BREAK_EVEN` with non-zero points, and confirms it stays excluded from `average_loss` in the range summary (contributing only to `final_result`).

Also fixed the other three findings (weekend-skip in `run_range`, rejecting "today" until session close, frontend Paris-local date) in a separate commit. Thanks for the thorough review!

---

api/services/backtest_service.py
```diff
@@ -0,0 +1,386 @@
...
+        take_profit_level = h1_high - TAKE_PROFIT_OFFSET_POINTS
+
+        for candle in candles:
+            candle_date = _candle_date(candle)
+            if position is None:
+                if not breached:
+                    if candle.lower < h1_low:
+                        breached = True
+                    continue
+                if candle.close >= h1_low:
```
RNiveau: **No invariant that the take-profit sits above the entry — a trade can be "born" already past its target and exit as a `take_profit` that's actually flat/negative.**

`take_profit_level = h1_high - 10` is fixed from the H1 range and never checked against `entry_price`. But `entry_price` is the *close of the reversal candle* (`candle.close >= h1_low`), which is only bounded from **below** by `h1_low` — nothing stops it closing above `h1_high - 10` (a strong reversal candle, or a narrow H1 range where `h1_high - h1_low < 10`).

When `entry_price >= take_profit_level`, the very next candle almost certainly satisfies `candle.higher >= take_profit_level`, so the position exits immediately — recorded as `TAKE_PROFIT` with `points = exit - entry` that can be **zero or negative**. In `_build_summary` a negative-points trade is then classified as a *losing* position while still carrying exit reason `take_profit`, which is contradictory, and it fires without any favorable move having occurred.

The spec doesn't cover "entry above the TP level," so this is undefined behavior rather than a spec violation, but it's reachable on volatile days and there's no test for it. Options: guard entry (skip/deem invalid when `entry_price >= take_profit_level`), or clamp so an at/above-target entry can't record a `take_profit`. At minimum, add a test pinning down the intended behavior.

---

frontend/src/components/BacktestDayDetail.tsx
```diff
@@ -0,0 +1,92 @@
...
+                </tr>
+              ))}
+            </tbody>
+          </table>
+        </>
+      )}
+
+      {detail.candles.length > 0 && (
+        <details className="backtest-candles-details">
+          <summary>5-minute candles from 10:00 ({detail.candles.length})</summary>
```
RNiveau: **Times are rendered as raw UTC and DST-shifted from the Paris-local frame the whole strategy is defined in — this undercuts the detail view's purpose.**

`trade.entry_time`, `trade.exit_time` and `candle.date` (lines 48/50/78) are the backend's naive-UTC datetimes, serialized as ISO strings and printed verbatim. The backend deliberately did all its window math in UTC (correct), but nothing converts back to Europe/Paris for display.

So on a summer (CEST) day, this section labeled "5-minute candles from 10:00" actually lists rows starting at `...T08:00:00`, and a 10:05 Paris entry shows as `07:05` / `08:05`. FR-015 and SC-004 exist specifically so a trader can "visually confirm the H1 reference levels and entry/exit points against the actual chart" — but their chart is in Paris local, so every timestamp here is off by 1–2h depending on DST, making the audit view actively misleading.

Suggest formatting all three timestamps in `Europe/Paris` (e.g. `toLocaleString('fr-FR', { timeZone: 'Europe/Paris' })`) and dropping the raw `T`. Same applies to the H1 window label wording.

---

RNiveau (commented):
## Second review — deeper pass (engine edge cases + presentation layer)

Went back over the parts I only skimmed the first time: the DI wiring, `model/__init__` exports and `UnitTime.M5`, the router + `get_candles_in_window` tests, and the frontend service/detail component. DI (`get_backtest_service` reuses the cached `get_candles_service` singleton), exports, and router validation coverage (invalid format, future date, end-before-start, unknown definition — all tested on both `/day` and `/run`) are all solid. `get_candles_in_window`'s naive-UTC assumption is confirmed by its own tests.

Two **new** findings beyond the first review, both left as inline comments:

1. **Take-profit isn't guaranteed to be above the entry** (`backtest_service.py:313`). `take_profit_level = h1_high − 10` is never compared to `entry_price`, and entry (the reversal candle's close) is only bounded below by `h1_low`. When entry lands at/above `h1_high − 10` (strong reversal candle, or a narrow H1 range), the position exits on the next candle as a `TAKE_PROFIT` with zero/negative points — which then classifies as a *losing* position while labeled take-profit. Undefined in the spec, reachable on volatile days, untested.

2. **Detail view shows UTC, DST-shifted times** (`BacktestDayDetail.tsx:63`). Entry/exit/candle timestamps are the backend's naive-UTC values printed raw, so the "5-minute candles from 10:00" list actually starts at 08:00 in summer, and entries display 1–2h off. Since FR-015/SC-004's whole point is visual verification against a Paris-local chart, this makes the audit view misleading. Format in `Europe/Paris`.

### Recap of severity
Everything found across both reviews is minor-to-medium and non-blocking on its own; the engine is faithful and well-tested. Ranking what I'd fix before shipping:

- **Blocker (already flagged by the author):** run against **real** FRA40.I data — SC-002's ≥6 hand-verified days. Nearly all the remaining risk lives in candle-boundary/DST/gap behavior that `MockSaxoClient` can't exercise.
- **Medium:** #2 above (misleading times in the verification view) and #1 (take-profit-below-entry) — both directly touch result correctness/interpretability.
- **Minor:** the break-even-gap points invariant, `run_range` sequential/weekend fetches, and the "today" incomplete-day gap from review 1.

No approve/request-changes — leaving as comments for the author to triage.

---

frontend/src/components/BacktestDayDetail.tsx
RNiveau: Fixed — added `formatParisTime` (`frontend/src/utils/backtestTime.ts`), which forces UTC parsing (appends `Z`) then formats via `Intl.DateTimeFormat(..., { timeZone: 'Europe/Paris' })`. Wired into entry time, exit time, and candle time, plus relabeled the table headers/summary to make the timezone explicit. Verified in a running browser with mocked API data: UTC `08:00`/`08:05` now correctly renders as Paris `10:00`/`10:05`.

---

api/services/backtest_service.py
RNiveau: Really good catch, resolved per explicit direction: take the trade only if the close-back is within 20pts of the H1 low *and* still below the take-profit level.

Added `_is_valid_entry(entry_price, h1_low, take_profit_level)` — `entry_price - h1_low <= 20 and entry_price < take_profit_level` — gating the entry in `_evaluate_trades`. A signal that satisfies FR-006 but fails either bound no longer opens a trade; the `breached` flag resets so the strategy immediately resumes searching for a fresh breakdown rather than sitting out the rest of the day.

Documented as new FR-006a plus an Edge Case bullet in `spec.md`, and updated the `data-model.md` state diagram. Added 9 new unit tests (`TestIsValidEntry` for the boundary logic directly, `TestEvaluateDayEntryValidityRule` for the end-to-end behavior including the fresh-signal-after-rejection case) — 100% coverage maintained on `backtest_service.py`.

---

api/services/backtest_service.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def evaluate_day(
+        self, definition: BacktestDefinition, trading_date: datetime.date
+    ) -> DayResult:
+        """Run the "CAC40 Bougie de 9h" rules for a single trading day."""
+        h1_start_utc, h1_end_utc = paris_reference_window_utc(trading_date)
+        h1_candle = self._fetch_h1_reference_candle(
+            definition.instrument, h1_start_utc, h1_end_utc
+        )
+        if h1_candle is None:
```
RNiveau, line 164: **`DayResult.candles` is stored in the raw fetch order (newest-first), so the detail view lists the 5-minute candles in reverse chronological order.**

Line 134 sorts into `chronological` for the engine, but here the DayResult is populated with the *unsorted* `five_min_candles`. `get_candles_in_window` preserves the order returned by `get_historical_data`, and the real `SaxoClient` returns candles **newest-first** (`sorted(..., reverse=True)`, `client/saxo_client.py:539`, matching the repo-wide "index 0 = newest" convention noted in CLAUDE.md).

`CandleResponse`/`DayDetailResponse` preserve that order, and `BacktestDayDetail.tsx` renders `detail.candles.map(...)` as-is — so the table under the heading **"5-minute candles from 10:00"** actually starts at ~17:25 and ends at 10:00, i.e. fully reversed. FR-015 calls for "the sequence of 5-minute candles **from 10:00 onward**," so this contradicts the spec and the label.

Note this is display-only — the engine is unaffected because it operates on `chronological` — which is exactly why the existing tests miss it (`test_traded_day_with_trades_and_candles` feeds candles in a fixed order and just checks they round-trip; nothing asserts chronological ordering of `DayResult.candles`).

Fix: store `candles=chronological` (or `sorted(...)`) in the DayResult, and add an assertion that the serialized candles are strictly time-ascending.

---

api/services/backtest_service.py
```diff
@@ -0,0 +1,406 @@
...
+                if not breached:
+                    if candle.lower < h1_low:
+                        breached = True
+                    continue
+                if candle.close >= h1_low:
+                    if self._is_valid_entry(
+                        candle.close, h1_low, take_profit_level
+                    ):
+                        position = _OpenPosition(
+                            entry_time=candle_date,
```
RNiveau, line 367: **Minor: an entry on the final 5-minute candle produces a zero-point `END_OF_DAY` trade that then classifies as a loss.**

If the reversal/entry fires on `candles[-1]`, the loop ends with a position still open and this block closes it at `last_candle.close`, which *is* the entry candle's close — so `points == 0`, `exit_reason == END_OF_DAY`. In `_build_summary` a non-BE zero-point trade counts as a losing position (the spec's documented "exactly-0-via-non-BE ⇒ loss" rule), so a trade that never had a chance to move records as a loss.

Edge, low frequency, and arguably acceptable, but if it's not intended you may want to suppress a same-candle entry+EOD close (no post-entry candle ⇒ no trade). Worth a one-line test either way to pin the intended behavior.

---

RNiveau (commented):
## Third review — data/presentation ordering + a boundary edge

This pass focused on data flow and ordering (how candles travel from `SaxoClient` → `get_candles_in_window` → `DayResult` → API → UI) and the loop boundaries. One real bug and one minor edge, both inline:

1. **Reverse-ordered candle table** (`backtest_service.py:144`). `DayResult.candles` keeps the raw newest-first fetch order while trades are computed on a separately-sorted `chronological` list. Because the real `SaxoClient` returns candles newest-first (`saxo_client.py:539`), the detail view's "5-minute candles from 10:00" table renders fully reversed (17:25 → 10:00), contradicting FR-015. P&L is unaffected (engine uses `chronological`), which is why the tests don't catch it. Fix: store `chronological` in the `DayResult` and assert time-ascending order.

2. **Zero-point EOD entry on the last candle** (`backtest_service.py:335`) records as a loss. Low-frequency boundary case; worth a test to pin intended behavior.

### Where this leaves the PR after three passes
The core rule engine remains correct and faithfully spec'd — the issues found across all three reviews sit in the layers *around* it (presentation, aggregation-time semantics, and a couple of un-guarded engine edges). Consolidated priority:

- **Blocker (author-flagged):** validate against real FRA40.I data — SC-002's ≥6 hand-verified days. Still the one true gate.
- **Should-fix before shipping:** the reversed candle table (#1 here) and the UTC/DST-shifted timestamps (review 2) — together they make the FR-015/SC-004 verification view show candles both in the wrong order *and* the wrong time, which defeats its purpose; plus take-profit-below-entry (review 2).
- **Minor / document-or-test:** break-even-gap points invariant (review 1), `run_range` sequential+weekend fetches (review 1), "today" incomplete-day acceptance (review 1), and the zero-point EOD edge (#2 here).

Still leaving all three reviews as **COMMENT** — nothing here is a correctness-of-P&L defect, but #1 + the review-2 time bug together are a strong case for tidying the detail view before this is used to judge the strategy. Author to triage.

---

api/services/backtest_service.py
RNiveau: Real bug, fixed — `DayResult.candles` now stores `chronological` instead of the raw, unsorted `five_min_candles`. Added a regression test that feeds candles newest-first (matching the real `SaxoClient`) and asserts the returned `DayResult.candles` come back time-ascending.

---

api/services/backtest_service.py
RNiveau: Keeping current behavior — this is exactly the case spec.md's Assumptions section already calls out explicitly ("an end-of-day close that happens to land on the entry price without break-even ever arming... is classified as a losing position, since only break-even-mechanism exits are labeled 'BE'"), so no code change needed here.

Added the one-line pinning test you suggested: `test_entry_on_last_candle_produces_zero_point_end_of_day` confirms the same-candle entry+EOD case (`points == 0`, `exit_reason == END_OF_DAY`), and `test_zero_point_end_of_day_trade_counts_as_losing` confirms `_build_summary` classifies it as a loss, not a win or BE.

---

api/models/backtest.py
```diff
@@ -0,0 +1,162 @@
...
+class DayDetailResponse(BaseModel):
+    date: datetime.date
+    status: str
+    h1_high: Optional[float] = None
+    h1_low: Optional[float] = None
+    candles: List[CandleResponse] = []
+    trades: List[TradeResponse] = []
+
+    @classmethod
+    def from_day_result(cls, day_result: DayResult) -> "DayDetailResponse":
```
RNiveau: I don't like this kind of method directly in the class. Do a util in place of that and go through the pr to remove other things like this

RNiveau: Fixed — replaced all six `@classmethod from_x(cls, x)` converters with plain module-level functions in `api/models/backtest.py`. The Pydantic classes are now pure data containers; `backtest_definition_to_response`, `day_result_to_response`, and `backtest_run_result_to_response` are called from the router, with `_trade_to_response`/`_candle_to_response`/`_day_result_summary_to_response`/`_backtest_summary_to_response` kept private since they're only used internally by the other conversion functions in that file.

Audited the rest of this PR's diff (`git diff` against the merge-base with `main`) for the same pattern — `api/models/backtest.py` was the only file in this feature with it, so nothing else to change.

100% coverage maintained on `api/models/backtest.py` and `api/routers/backtest.py`.

---

tests/api/routers/test_backtest.py
```diff
@@ -0,0 +1,321 @@
+import datetime
```
RNiveau, line 1: This file test is useless, it tests mock. Remove it

RNiveau: Checked with you before deleting outright — the fixture pattern here (mocked `BacktestService` via `TestClient` + `dependency_overrides`) is the same one `test_watchlist.py`/`test_workflow.py` use for `SaxoClient`/`DynamoDBClient`, and it exercises real router logic (date validation, unknown-definition handling, response serialization), not just mock call assertions. You asked to trim it rather than remove it entirely.

Cut 3 of 13 tests: the NO_TRADE/NO_DATA-only day-detail variants (redundant with the combined trades+candles happy path) and the range endpoint's duplicate future-date check (same `_parse_date` path already covered by the day endpoint's test). Kept one representative 200 per endpoint plus every distinct validation/error path (bad format, future date, today-not-closed, end-before-start, unknown definition) — still 100% coverage on `api/routers/backtest.py` and `api/models/backtest.py`.

---

api/services/backtest_service.py
```diff
@@ -0,0 +1,406 @@
...
+# hardcoded implementation, not a generic configurable engine.
+STOP_LOSS_POINTS = 50
+TAKE_PROFIT_OFFSET_POINTS = 10
+BREAK_EVEN_TRIGGER_POINTS = 20
+MAX_ENTRY_DISTANCE_FROM_LOW = 20
+FIVE_MINUTE_HORIZON = 5
+H1_HORIZON = 60
+
+
+def paris_reference_window_utc(
```
RNiveau, line 56: Why don't you use Market.EU ?

RNiveau: Checked `EUMarket`/`market_in_utc()` before writing this — it's close, but not a safe drop-in for both boundaries this function needs:

- **9:00 start** matches `EUMarket.open_hour=9, open_minutes=0` exactly.
- **17:30 session close** doesn't: `EUMarket.close_hour=17` (confirmed via `tests/utils/test_helper.py:1018-1020` — summer→15 UTC, winter→16 UTC, i.e. 17:00 Paris local). `Market` has no `close_minutes` field, so it can't express the actual Euronext Paris close of 17:30 I need for FR-008's end-of-day exit — `EUMarket`'s 17:00 looks like a deliberate rounding for its own callers (H4/daily candle bucketing in `utils/helper.py`), not the literal session close.
- `market_in_utc()` is also shaped specifically around `Market.open_hour`/`close_hour` (whole-hour only) — it doesn't give me the 10:00 boundary either, which isn't a session open/close at all, just where 5-min evaluation starts.

So a straight swap would silently shrink my evaluation window by 30 minutes (17:30 → 17:00) — a real behavior change, not just a refactor. I kept `PARIS_TZ = ZoneInfo("Europe/Paris")` + explicit `datetime(...)` construction instead.

If you'd rather `EUMarket` be the single source of truth for Euronext Paris hours (i.e. fix its close to 17:30 and extend `Market`/`market_in_utc` to support close minutes), I can do that — but that's a shared-utility change touching the H4/daily candle building it's used for elsewhere, so wanted to flag the tradeoff rather than change it unilaterally. Let me know if you want that instead.

---

RNiveau: Done — added `Market.end_minute: int = 0` (`model/__init__.py`): `EUMarket.end_minute=30` (real Euronext Paris close 17:30, 30 min past its `close_hour=17` label) and `USMarket.end_minute=60` (real NYSE close 16:00, a full hour past its `close_hour=15` label). Left `close_hour` untouched on both since `build_h4_candles_from_h1`/`build_daily_candles_from_h1` rely on its current "last full H1-candle label" meaning.

`market_in_utc` (`utils/helper.py`) now carries `end_minute` through unchanged in the returned `Market` (same as `open_minutes`/`h4_blocks` — DST shifts are whole hours, so minutes never move).

`backtest_service.py`'s `paris_reference_window_utc`, `paris_session_end_utc`, and `is_today_not_yet_closed` now derive their times from `EUMarket()` via `market_in_utc` instead of a second hand-rolled Paris conversion. Verified output is identical to before (existing DST summer/winter test cases for the reference window and session end still pass unchanged) — 394 tests pass, `backtest_service.py` still 100% covered, extended `test_market_in_utc` to assert `end_minute` round-trips correctly. Also updated `research.md` §1, which had noted `EUMarket` wasn't DST-aware at the time this feature was first written — that's no longer true now that `market_in_utc` exists.

---

api/services/backtest_service.py
```diff
@@ -0,0 +1,421 @@
...
+        d.month,
+        d.day,
+        market.close_hour,
+        market.end_minute,
+        tzinfo=PARIS_TZ,
+    )
+    return current < session_end_local
+
+
+def _candle_date(candle: Candle) -> datetime.datetime:
```
RNiveau, line 115: this will raise an exception when date is None, it's a bit rough

RNiveau: Fair — `assert` is also stripped entirely under Python's `-O` flag, so it wasn't just a rough exception, it wasn't even guaranteed to run.

Replaced with an explicit `if candle.date is None: raise SaxoException(...)`, which mypy narrows the same way for the type checker but always executes and raises a properly-named exception instead of a bare `AssertionError`. It's still unreachable in practice today (`get_candles_in_window` filters out dateless candles before `_candle_date` ever sees them), but now it fails predictably if that ever changes. Added tests for both the happy path and the error path — 100% coverage maintained.

---

api/services/backtest_service.py
```diff
@@ -0,0 +1,421 @@
...
+        take_profit_level = h1_high - TAKE_PROFIT_OFFSET_POINTS
+
+        for candle in candles:
+            candle_date = _candle_date(candle)
+            if position is None:
+                if not breached:
+                    if candle.lower < h1_low:
+                        breached = True
+                    continue
+                if candle.close >= h1_low:
```
RNiveau: I think we're bad here, the plan is:
- we need a candle bellow the h1 low (good with line 340)
- then we need a close above h1 low (good here)
- then we need a breakout of this candle, I don't see it

RNiveau: You're right, real gap — confirmed and implemented per your clarification: entry now requires a later candle's high to break above the reversal candle's high, not just the close-back itself.

`_evaluate_trades` now tracks a pending `candidate` (the reversal candle) instead of entering immediately:
- if a later candle's high breaks above `candidate.higher` → entry at `candidate.higher`, or that candle's open if it gapped above it
- if a later candle closes back below the H1 low first → the candidate is discarded and that same candle counts as a fresh breach (its low is necessarily below the H1 low too)
- otherwise (still above H1 low, hasn't broken out yet) → that candle becomes the new candidate, rolling the breakout level forward

The 20pt/take-profit validity check now runs against this actual fill price rather than the old candle-close.

Rewrote the affected tests for the extra candle this needs, and added `TestBreakoutConfirmation` covering the rolling-candidate, close-below-reset, and gap-fill-on-entry cases directly. Updated spec.md (FR-006 now describes the candidate, new FR-006b for the breakout confirmation, FR-006a/FR-007 updated to reference the real entry price) and data-model.md's state diagram. 400 tests pass, `backtest_service.py` still 100% covered.

---

PR #632, opened 2026-07-17
fix: make market session hours DST-aware via central helper

RNiveau:
Exchange sessions are defined in local wall-clock time (Euronext Paris
09:00-17:00, NYSE 09:30-15:00), but Saxo returns candle timestamps in
UTC. The Market models hardcoded the summer-UTC hours (EU 7-15, US
13-19), so every hour-matching comparison in the candle builders drifted
by one hour in winter (CET/EST), breaking H1/H4/daily aggregation for
roughly half the year.

Store the session in exchange-local time plus an IANA timezone on Market
and add a single helper, market_in_utc(), that converts those hours to
UTC for a given reference date using stdlib zoneinfo. Apply it at the
three occurrences (build_h4_candles_from_h1, build_daily_candles_from_h1,
CandlesService._build_h1_from_30m), each deriving its reference date from
the data it already holds. Summer conversions reproduce the previous UTC
hours exactly, so existing tests are unchanged; winter dates now resolve
correctly.

Claude-Session: https://claude.ai/code/session_01LWuyypCv3WkEHC3SM1VDtx

### DIFF
--- a/utils/helper.py
+++ b/utils/helper.py
@@ -58,15 +58,6 @@ def market_in_utc(market: Market, reference: datetime.datetime) -> Market:
     )
 
 
-def _market_reference_date(
-    candles: List[Candle],
-) -> datetime.datetime:
-    """Pick the DST reference date for a batch of candles (newest first)."""
-    if candles and candles[0].date is not None:
-        return candles[0].date
-    return get_date_utc0()
-
-
 def build_current_weekly_candle_from_daily(
     candles: List[Candle],
 ) -> Optional[Candle]:
@@ -124,24 +115,28 @@ def get_date_utc0() -> datetime.datetime:
     return datetime.datetime.now(tz=datetime.UTC)
 
 
-def build_h4_candles_from_h1(
-    candles: List[Candle], market: Market
-) -> List[Candle]:
-    market = market_in_utc(market, _market_reference_date(candles))
+def _h4_ending_hours(market: Market) -> dict:
+    """Map each H4 block's UTC ending hour to its size for `market`."""
     ending_hours = {}
     cumulative = 0
     for block_size in market.h4_blocks:
         cumulative += block_size
-        ending_hour = market.open_hour + cumulative - 1
-        ending_hours[ending_hour] = block_size
+        ending_hours[market.open_hour + cumulative - 1] = block_size
+    return ending_hours
 
+
+def build_h4_candles_from_h1(
+    candles: List[Candle], market: Market
+) -> List[Candle]:
     candles_h4 = []
     i = 0
     while i < len(candles):
         candle_date = candles[i].date
         if candle_date is None:
             i += 1
-        elif candle_date.hour in ending_hours:
+            continue
+        ending_hours = _h4_ending_hours(market_in_utc(market, candle_date))
+        if candle_date.hour in ending_hours:
             block_size = ending_hours[candle_date.hour]
             if i + block_size - 1 >= len(candles):
                 break
@@ -160,8 +155,6 @@ def build_h4_candles_from_h1(
 def build_daily_candles_from_h1(
     candles: List[Candle], market: Market
 ) -> List[Candle]:
-    market = market_in_utc(market, _market_reference_date(candles))
-    ending_hour = market.close_hour - (1 if market.open_minutes == 30 else 0)
     num_h1 = (
         market.close_hour
         - market.open_hour
@@ -174,7 +167,12 @@ def build_daily_candles_from_h1(
         candle_date = candles[i].date
         if candle_date is None:
             i += 1
-        elif candle_date.hour == ending_hour:
+            continue
+        utc_market = market_in_utc(market, candle_date)
+        ending_hour = utc_market.close_hour - (
+            1 if market.open_minutes == 30 else 0
+        )
+        if candle_date.hour == ending_hour:
             if i + num_h1 - 1 >= len(candles):
                 break
             candles_daily.append(
--- a/utils/helper.py
+++ b/utils/helper.py
@@ -1,8 +1,64 @@
 import datetime
 from typing import List, Optional
+from zoneinfo import ZoneInfo
 
 from model import Candle, Market, UnitTime
 from utils.logger import Logger
+
+
+def market_in_utc(market: Market, reference: datetime.datetime) -> Market:
+    """Convert a market's exchange-local session hours to UTC, DST-aware.
+
+    The Saxo API returns candle timestamps in UTC, while an exchange's
+    session is defined in its local wall-clock time (Euronext Paris trades
+    09:00-17:00 local all year, NYSE 09:30-15:00 local). Because the UTC
+    offset of that local time changes with daylight saving, the session's
+    UTC hours shift by one hour between summer and winter. Centralising the
+    conversion here keeps every candle builder DST-correct instead of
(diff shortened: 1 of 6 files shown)


RNiveau (commented):
## Review — DST-aware market session hours

Clean, well-targeted fix. `Market` now stores exchange-local hours + an IANA timezone, and a single `market_in_utc()` helper converts to UTC per reference date via stdlib `zoneinfo`, applied at the three aggregation sites. I verified the DST math independently: summer reproduces the old hardcoded UTC hours **exactly** (EU 7–15, US 13–19), winter shifts +1 (EU 8–16, US 14–20), so existing tests stay green and winter is now correct. Every caller (`snapshot.py`, `workflow_engine.py`, `alerting.py`, `build_weekly_candles`) passes the market *through* to the builders rather than reading `.open_hour`/`.close_hour` directly, so moving to local hours breaks no other call site. Conversion is single-pass and idempotent. Nice design.

### Issues & risks

**1. `zoneinfo` on the slim Lambda image — deployment risk (high).** Runtime image is `python:3.12-slim-bullseye` (`Dockerfile:11`). Debian *slim* images don't install the `tzdata` OS package, so `/usr/share/zoneinfo` may be absent — and `tzdata` is **not** a Linux dependency here (`poetry.lock` only pulls it in on Windows). `ZoneInfo("Europe/Paris")` tries system tzdata, then falls back to the `tzdata` PyPI package; with neither present it raises `ZoneInfoNotFoundError` at runtime, on every candle build. Recommend adding `tzdata` to Poetry main deps (base-image-independent guaranteed fallback), or `apt-get install -y tzdata` in the Dockerfile. Worth confirming against the deployed image before merge.

**2. One DST regime per *batch* — accuracy gap for large `count` (medium).** `market_in_utc` is called once per batch using a single reference date (the newest candle). For a batch spanning a DST transition, the whole batch is converted with the newest candle's regime, so opposite-season candles are mis-filtered:
- ref = winter → summer's 07:xx open H1 fails `hour >= open_hour(8)` and is dropped.
- ref = summer → winter's 16:00 close H1 fails `hour <= close_hour(15)` and is dropped.

Dropping a boundary H1 makes that day's `num_h1` count not match, so the daily/H4 candle for those days is skipped. Short windows (MA50 ≈ 55 candles) usually stay within one season, but `COMBO` requests `count=750` daily candles (~3 years), which definitely spans many transitions — so historical indicator values are subtly off in the season opposite to "now". Still a net improvement (recent candles are always correct now), but the fully-correct fix is per-candle conversion (each candle's own date selects its regime) rather than per-batch. At minimum, worth documenting the limitation.

**3. `to_utc_hour` returns only `.hour` — latent constraint (low).** It discards date rollover, and `build_daily_candles_from_h1` then computes `num_h1 = close_hour - open_hour + …`. Fine for EU/US (session within one UTC day, open < close), but a future market whose local→UTC conversion crosses midnight would yield a wrong/negative count. A short note or an `open < close` assertion would help.

### Test coverage

`test_market_in_utc` (both markets × both seasons) and `test_build_daily_candle_from_h1_winter_dst` directly pin the regression — good picks. Gaps: no test for the `_build_h1_from_30m` conversion site, no DST **transition-day** test, and no cross-season batch test to characterize issue #2. A winter H4 test would mirror the winter daily test.

### Minor

- The `raise SaxoException(f"No data returned for {code}")` reformat is unrelated to DST — harmless scope creep.
- `market_in_utc` returns a base `Market`, not the `EUMarket`/`USMarket` subclass — fine since only fields are read.
- Docstrings are clear and accurate.

### Verdict

Solid, correct fix for the common case. I'd treat **#1 (tzdata)** as a blocker to confirm, and **#2** as a documented known-limitation (or fix per-candle) rather than block on. Nothing else is serious.

---

RNiveau:
Thanks for the thorough review — all three points addressed:

**#1 (tzdata blocker):** Confirmed real — the runtime image is `python:3.12-slim-bullseye` (no OS tzdata) and the `tzdata` PyPI package was Windows-only in the lock, so `ZoneInfo("Europe/Paris")` would raise `ZoneInfoNotFoundError` at runtime. Added `tzdata` to the main Poetry deps (base-image-independent `zoneinfo` fallback per PEP 615); it lands in the `main` group so the Dockerfile's `poetry install --only main` picks it up. Verified `ZoneInfo("Europe/Paris")` resolves via the PyPI package. (`d2a6864`)

**#2 (per-batch regime):** Fixed properly rather than documented — went with per-candle conversion. `build_h4_candles_from_h1`, `build_daily_candles_from_h1`, and `CandlesService._build_h1_from_30m` now compute the UTC session hours from **each candle's own date**, so a batch spanning DST transitions (e.g. `COMBO`'s ~750 daily candles) aggregates every day under its own regime. The intra-day count (`num_h1`) is DST-invariant so it's derived once from the raw local hours. Added a cross-season batch test (a winter + a summer day in one call, which the old per-batch code would have dropped one of) and a winter H4 test. (`1175505`)

**#3 (hour-only conversion):** Added a `Note` to `market_in_utc` documenting the single-UTC-day / `open < close` assumption (holds for Euronext + US). (`d2a6864`)

On the minor points: the `raise SaxoException(...)` reformat was `black` collapsing a multi-line call — reverting it would fail `black --check`, so I left it. `market_in_utc` intentionally returns a base `Market` since only fields are read.

All 38 helper/candles tests green; summer results are unchanged (each June/July date resolves to the same UTC hours as before).

---

RNiveau (commented):
All three points resolved — and #2 fixed properly with per-candle regime selection rather than documented. Traced the new logic:

- **tzdata:** added to `main` Poetry deps + `poetry.lock` (`groups = ["main"]`), so `poetry install --only main` bundles it for the slim image. ✅
- **Per-candle DST:** `num_h1` derived once from raw local hours (DST-invariant), `ending_hour`/`_h4_ending_hours` recomputed per candle from each candle's own date. Winter/summer candles at the same hour now disambiguate by date. ✅
- **Assumption documented** in `market_in_utc`. ✅

The cross-season batch test is exactly the regression the per-batch code would have failed, and summer results are unchanged (no regression). LGTM — ready to merge once CI is green. Nice work.

_(Would've hit Approve, but GitHub won't let the authoring account approve its own PR.)_

---

utils/helper.py
```diff
@@ -1,10 +1,63 @@
...
+        within a single UTC day with ``open_hour < close_hour`` (true for
+        Euronext and US exchanges). A market whose local->UTC conversion
+        crosses midnight would need day-aware handling.
+    """
+    tz = ZoneInfo(market.timezone)
+    if reference.tzinfo is None:
+        reference = reference.replace(tzinfo=datetime.UTC)
+    local_date = reference.astimezone(tz).date()
+
+    def to_utc_hour(hour: int, minute: int) -> int:
```
RNiveau: I hate this method in method pattern. Fix it please

RNiveau:
Addressed the review comment on `utils/helper.py` (method-in-method) in `6732df8`: extracted the nested `to_utc_hour` closure into a module-level `_local_hour_to_utc(local_date, hour, minute, tz)` helper, so `market_in_utc` no longer defines a function inside itself. 38 tests green, black/flake8/mypy clean.

(Couldn't reply on the review thread directly — there's a pending review open on the PR, so GitHub blocks new review comments until it's submitted.)

---

PR #645, opened 2026-07-21
feat: add short positions to CAC40 Bougie de 9h backtest (spec 021)

RNiveau:
Extend the hardcoded "CAC40 Bougie de 9h" backtest to trade both
directions symmetrically. The long side (reversal off the H1 low) is
unchanged; a new short side trades the mirror-image reversal off the
H1 high: break above the high, reversal candle closing back at/below
it, breakdown confirmation below that candle's low, with stop-loss at
entry+50, take-profit at H1 low+10, and break-even armed at entry-20.

Both directions are evaluated on every day with at most one position
open at a time (either side); whichever confirms first holds the floor.
Trades now carry a direction (reusing model.enum.Direction), exposed in
the API and shown as long/short in the day-detail view. Points are
signed P&L per direction so the aggregate win/loss/BE figures are
unchanged.

Updates spec 021 (clarifications, FR-017–FR-022, edge cases),
data-model, plan, and the API contract accordingly.

Claude-Session: https://claude.ai/code/session_01DmkTc7rmHbmKHQ5d7fCKuF


RNiveau:
## Code review

Solid, well-tested change. I traced the new logic against the previous inline engine and the spec — no correctness bugs found. Below is a summary.

### What it does
Extends the "CAC40 Bougie de 9h" backtest to trade both directions symmetrically: the long side (reversal off the H1 low) is unchanged, a new short side mirrors it around the H1 high. Both directions are searched concurrently while flat, but at most one position is open at a time (long-precedence tiebreak). `Trade` now carries a `direction` (reusing `model.enum.Direction`), surfaced via the API, CSV export, and the day-detail UI, with spec 021 updated to match.

### Strengths
- **Good refactor** — extracting `_DirectionSearch` (`feed()`/`reset()`) and `_resolve_exit()` keeps the two-direction engine readable. The long branch is a faithful 1:1 port of the original inline state machine.
- **Mirror logic verified** — short stop (`entry+50`), TP (`h1_low+10`), BE arm (`low <= entry-20`), gap-fill on both exits, and signed points (`entry-exit`) are all correct reflections of the long side. Hand-computed test expectations check out.
- **Tricky spots preserved for both directions** — same-candle stop-before-take-profit priority (FR-009), BE never armed on a candle that already exits, and entry/closing candles correctly excluded from re-evaluation. The two `_DirectionSearch` instances share no mutable state.
- **Strong test coverage** — short-side mirror of the long suite plus the key cross-direction cases (`test_long_then_short_sequential_same_day`, `test_high_breach_closing_a_long_does_not_also_open_a_short`).

### Minor suggestions (non-blocking)
- **`Trade.direction = Direction.BUY` default**: only serves backward-compat; a future site that forgets it would silently mislabel a short as a long rather than fail. Reasonable tradeoff as documented, just noting it.
- **Frontend `formatDirection`**: `direction === 'Buy' ? 'long' : 'short'` hardcodes the enum's serialized value and treats any non-`'Buy'` as short. Fine for a two-value enum; noting the string coupling to `Direction.value`.
- **CSV**: `direction` inserted mid-record (`exit_reason,direction,points`) — fine for internal use, would break any positional external parser.
- **Test gaps to consider**: no short-side gap-fill *entry* test (confirming candle opens below the candidate low), and no direct short tests for FR-020b (candidate discarded by close back above the high) / FR-020c (candidate rolls forward). Low risk given the shared structure.
- Leftover "(added after PR review)" phrasing in `_is_valid_long_entry`'s docstring — minor cleanup.

Nothing here blocks merging.

---

api/services/backtest_service.py
```diff
@@ -125,16 +153,132 @@ def _candle_date(candle: Candle) -> datetime.datetime:
...
+        when this candle confirms a tradeable breakout, otherwise None
+        (updating the internal breach/candidate state)."""
+        if self.is_long:
+            return self._feed_long(candle)
+        return self._feed_short(candle)
+
+    def _feed_long(self, candle: Candle) -> Optional[float]:
+        if self.candidate is None:
+            if not self.breached:
+                if candle.lower < self.h1_low:
```
RNiveau: It should be close here. We want a confirmed candle outside the h1 range. Same for short position

RNiveau: Done in 56ef739. The breach now requires a candle that **closes** outside the H1 range — `candle.close < h1_low` for a long and `candle.close > h1_high` for a short — instead of an intrabar wick. A candle that pierces the range but closes back inside no longer arms a breach. This also makes the initial breach consistent with the discard/re-breach rule (FR-006b/FR-020), which already keyed off the close. Spec 021 (new clarification + FR-006/FR-006b/FR-020 + edge cases) and the data-model state transitions are updated, and I added wick-that-closes-inside no-trade tests for both directions.

---

PR #647, opened 2026-07-21
feat: parametrize backtest strategy thresholds

RNiveau:
Make the four "CAC40 Bougie de 9h" thresholds tunable instead of
hardcoded: stop-loss, take-profit offset, break-even trigger, and max
entry distance. Defaults reproduce the previous hardcoded values, so an
unparametrized run is unchanged.

- model: add BacktestParameters dataclass with the defaults
- service: thread params through evaluate_day/run_range and the trade
  engine (_OpenPosition, _DirectionSearch, _resolve_exit)
- api: expose the four values as optional, positive query params on the
  day/run and CSV endpoints
- frontend: add a Parameters panel to the Backtest page and pass the
  applied values through runs, drilldowns, and CSV exports
- tests: cover default reproduction, threshold overrides, param
  passthrough, and validation

Claude-Session: https://claude.ai/code/session_01VCq7ek2XCZukGNHP6g51a4


RNiveau:
## Code review

### Overview
Makes the four "CAC40 Bougie de 9h" thresholds (stop-loss, take-profit offset, break-even trigger, max entry distance) tunable per run instead of module constants. A new `BacktestParameters` dataclass carries the values; they're threaded from four optional query params → service (`evaluate_day`/`run_range`) → trade engine (`_OpenPosition`, `_DirectionSearch`, `_evaluate_trades`, `_resolve_exit`), and surfaced in the frontend via a Parameters panel. Defaults reproduce the old hardcoded values, so unparametrized runs are unchanged.

Overall a clean, well-scoped, well-tested refactor. The threading is consistent and the backward-compat story (`params or BacktestParameters()`) is solid. A few points below.

### Issues

**1. `parameters` prop is dead in `BacktestDayDetail` (main finding)**
`frontend/src/components/BacktestDayDetail.tsx` destructures `parameters` but never uses it — it's added to the interface and passed from `Backtest.tsx` at all three call sites (day, drilldown, range drilldown), but never rendered. Either this is an **incomplete feature** (the intent was presumably to show which thresholds produced this day's result — useful in a drilldown), or it's **dead plumbing** to remove. Given the PR description ("pass the applied values through ... drilldowns"), it reads as intended-but-unfinished. Note: `noUnusedParameters` is on, but TS exempts destructured object-binding elements, so this won't fail `tsc` — it just silently does nothing.

**2. UI always sends all four params**
`initialParamValues()` pre-fills every field with its default (`"50"`, `"10"`, …), and `buildParameters()` includes any non-empty field. So a "default" run from the UI always sends `stop_loss_points=50&...` rather than an empty query. Functionally identical, but it contradicts the "byte-for-byte the original request" claim in the spec/contract — that only holds for programmatic callers, not the page. Minor.

**3. `0` is reachable from the input but rejected server-side with a generic error**
The number inputs use `min="0"`, which permits `0`, but the backend enforces `gt=0` and returns 422; the catch block then shows the generic `"Failed to run backtest"` with no hint that a threshold was invalid. Consider making `min` reflect the real constraint or mapping 422 to a clearer message. Minor UX.

### Strengths
- **Backend threading is correct and complete** — stop distance, entry-distance validity, take-profit levels, and break-even arming all consume the per-run value; both directions share params symmetrically.
- **Backward compatibility** is airtight: `params or BacktestParameters()` at both service entry points, `_DEFAULTS` shared between the free functions and the dataclass so "unparametrized" has one definition.
- **Test coverage** is behavior-focused: default reproduction, per-threshold overrides with concrete price math, max-entry-distance rejection, param passthrough, and 422 validation. The existing `run_range` mocks were correctly updated to the new `lambda d, date, params` signature.
- **Conventions**: uses the `Direction` enum, keeps validation at the API boundary (`Query(gt=0)`) rather than in the model, and all the 021 specs (spec/plan/data-model/contract/tasks) are updated — including retiring the old "hardcoded thresholds" Complexity-Tracking exception.

### Verdict
Solid PR. Worth resolving **#1** before merge — decide whether `BacktestDayDetail` should render the applied parameters or drop the prop. #2 and #3 are optional polish.

PR #650, opened 2026-07-22
feat: Ouinex crypto provider foundation and provider selection (spec 024 phases 1-3)

RNiveau:
## Summary

Implements **phases 1–3** of spec [024-ouinex-provider](specs/024-ouinex-provider/): the setup, foundational layer, and User Story 1 (Ouinex recognized as a selectable crypto provider). Ouinex is added as a sibling to Binance at each layer, coexisting with it. Search, indicators, and reporting (US2–US4) are intentionally **not** included — those remain gated on research items R1/R3/R5 (real Ouinex GraphQL schema confirmation) per `tasks.md`.

### Phase 1 — Setup
- `config.yml`: add `ouinex_graphql_url` (default `https://live-api.ouinex.com/graphql`)
- `secrets.yml.example`: new template documenting `ouinex_api_key` / `ouinex_secret_key` (and existing secrets), since `secrets.yml` is gitignored and no template existed

### Phase 2 — Foundational
- `model/enum.py`: add `Exchange.OUINEX` and an `Exchange.is_crypto()` helper (True for Binance/Ouinex) to avoid scattered `in (BINANCE, OUINEX)` checks
- `utils/configuration.py`: add `ouinex_keys` and `ouinex_graphql_url` properties
- `client/ouinex_client.py`: new `OuinexClient(key, secret, graphql_url)` with a GraphQL POST transport and private JWT sign-in / token-refresh lifecycle (`_ensure_token`, `_execute` with a one-shot 401 refresh-retry)
- `model/__init__.py`: centralize the "map as binance" identity in a `crypto_account()` factory; `BinanceReportService._get_binance_account` now delegates to it (single source of truth for both report services)
- `api/dependencies.py`: cached `get_ouinex_client()`
- `utils/exception.py`: add `OuinexException`

### Phase 3 — US1 (provider selection)
- `api/routers/fund.py`: `/api/fund/accounts` now also returns the `ouinex_main` pseudo-account (key `ouinex`, name `Ouinex`) alongside `binance_main`
- `api/routers/homepage.py`: map `exchange == "ouinex"` → `Exchange.OUINEX`

## Tests
- `tests/client/test_ouinex_client.py`: JWT sign-in, token reuse/refresh, and GraphQL transport (bearer header, 401 refresh-retry, error handling)
- `tests/model/test_enum.py`: `Exchange.get_value("ouinex")` resolution and `is_crypto()`
- `tests/api/routers/test_fund.py`: `/api/fund/accounts` includes both `ouinex_main` and `binance_main`

## Verification
- Full suite: **460 passed, 10 skipped** (no regressions)
- `black`, `isort`, `flake8`, `mypy` all clean on the changed files

## Notes
- The `OuinexClient` transport is built to the research doc's documented-but-unconfirmed (VERIFY) GraphQL/JWT shape (resolves R4 provisionally). The actual sign-in mutation and field names should be confirmed against the real Ouinex schema before the search/candles/report methods land in phases 4–6.
- Binance paths are unchanged behaviorally; the only Binance-side edit routes its pseudo-account through the shared `crypto_account()` factory.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ST1dwao7yrcydK61shAquo)_


tests/model/test_enum.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+from model.enum import Exchange
```
RNiveau: This file is useless, we're testing a hardcoded code

api/routers/homepage.py
```diff
@@ -57,11 +57,12 @@ async def get_homepage(
...
-                    Exchange.BINANCE
-                    if exchange_str == "binance"
-                    else Exchange.SAXO
-                )
+                if exchange_str == "binance":
+                    exchange = Exchange.BINANCE
+                elif exchange_str == "ouinex":
+                    exchange = Exchange.OUINEX
+                else:
+                    exchange = Exchange.SAXO
```
RNiveau, line 65: `exchange_str == "ouinex"` now resolves to `Exchange.OUINEX`, but `IndicatorService.get_asset_indicators()` only special-cases `Exchange.BINANCE`:

```python
if exchange == Exchange.BINANCE:
    return await self._get_binance_asset_indicators(code, unit_time)
else:
    return await self._get_saxo_asset_indicators(code, country_code, unit_time)
```

`Exchange.OUINEX` silently falls into the **Saxo** indicator path rather than an "unsupported yet" path. Likely dead code today (nothing tags a watchlist item `ouinex` yet), but once phase 4+ lands Ouinex watchlist entries, this will query Saxo's API with a crypto code — most likely a `SaxoException`, or worse, silently wrong data if the code collides with a real Saxo asset.

Since `Exchange.is_crypto()` was added in this same PR specifically to avoid this class of missed branch (per `data-model.md`), might be worth guarding this call site now (skip/log/raise for crypto exchanges other than Binance) rather than leaving the fallthrough for a future PR to remember.

---

RNiveau (commented):
Flagging one item from review — see inline comment on the homepage.py exchange mapping. Everything else (mypy fix aside, already called out separately) looked solid.

---

api/routers/homepage.py
RNiveau: Good catch — fixed in babbb08. `get_asset_indicators()` now guards with `exchange.is_crypto()`: Binance keeps its path, any other crypto exchange (i.e. `OUINEX` until the US3 candle path lands) raises a clear `SaxoException("Indicators are not supported yet for ouinex")` instead of falling through to the Saxo path. That surfaces as a clean 400 on the indicator route and is logged-and-skipped on the homepage, rather than querying Saxo with a crypto code.

---

PR #652, opened 2026-07-22
feat: add "Bougie de 9h (time cut)" backtest variant

RNiveau:
## Summary

Adds a second hardcoded backtest — **"CAC40 Bougie de 9h (time cut)"** (code `B9HTC`) — that runs the exact same scenario as the existing `B9H` strategy plus one new exit rule:

> When a position is open, once **30 minutes** have elapsed since entry, if it has **never moved more than 5 points** in its favor, it is closed at market (recorded as a new `time_cut` exit reason).

The two backtests appear side by side in the Backtest menu so their results can be compared. The original `B9H` is completely unchanged.

## Behavior details

- **Favorable move** = max favorable excursion since entry (long: highest high − entry; short: entry − lowest low), measured from the first candle after entry (the entry candle is intrabar, mirroring break-even arming).
- **Inclusive boundary**: "never been higher than 5 pts" means a best move of exactly 5.0 pts still gets cut (`≤ 5`); only a move strictly above 5 saves the position.
- **Fill**: the close of the first 5-minute candle at/after the 30-minute mark (not a price-level exit, so the gap-fill convention does not apply).
- **Exit ordering**: checked after stop-loss/break-even and take-profit (those win if reached on the same candle), and before the end-of-day fallback. A trade that reaches its stop/target before the 30-min mark, or is entered within 30 min of the close, is never time-cut.

## Changes

- `model/enum.py` — new `Strategy.B9HTC` and `ExitReason.TIME_CUT`.
- `model/backtest.py` — `BacktestDefinition` carries optional `time_cut_minutes` / `time_cut_min_favorable_points` (both `None` on plain `B9H`, so it is untouched).
- `api/services/backtest_service.py` — the new definition (30 min / 5 pts); `_OpenPosition` tracks max favorable excursion; a `_resolve_time_cut` check slots into `_resolve_exit`. The new definition surfaces automatically via `/api/backtest/definitions`; the `time_cut` reason flows through the existing response/CSV serialization unchanged.
- `tests/api/services/test_backtest_service.py` — 6 new cases (long/short cut, the >5 pt survival case, the inclusive `=5` boundary, and an isolation test proving plain `B9H` is never cut).
- `specs/021-backtest-menu-hardcoded/spec.md` — amended in place (matching the prior short-positions / parametrization update pattern): update-input line, a 2026-07-22 clarifications session, User Story 1c, edge cases, FR-028–FR-031, and updates to Key Entities, Success Criteria (SC-006), and Assumptions.

## Testing

- `pytest tests/api tests/services` — **299 passed, 5 skipped** (pre-existing skips).
- `black` / `isort` / `flake8` clean on the changed files; `mypy` clean (the one reported error is a pre-existing `aioboto3` stub issue in an unrelated file).

---
_Generated by [Claude Code](https://claude.ai/code/session_016EFe58t22DtxSXq6dVP1Zd)_


RNiveau (commented):
Reviewed the implementation end-to-end (entry-candle exclusion, exit ordering, deadline math, MFE tracking, long/short symmetry, and the frontend/CSV pass-through) — it's correct and well-tested. A few **minor, non-blocking** notes:

1. **Zero-point time-cut is classified as a loss** — `api/services/backtest_service.py:411`. Classification is `BREAK_EVEN → be`, `points > 0 → win`, `else → loss`, so a time-cut that fills exactly at entry (0 points) lands in `losing`. This matches existing `END_OF_DAY` behavior and is consistent with FR-031 ("sign of its points result"), but a 0-point result is genuinely sign-less. Inherited from the existing convention rather than introduced here — flagging for awareness, no change required.

2. **`int` vs `Optional[float]`** — `time_cut_min_favorable_points=5` is passed as an `int` where the field is annotated `Optional[float]` (both in the `B9HTC` definition and the `TIME_CUT_DEFINITION` test fixture). Harmless and mypy-clean; `5.0` would match the annotation. Cosmetic.

3. **Side-effect inside a predicate-style `@staticmethod`** — `_resolve_time_cut` mutates `position.max_favorable_points` as a side effect of what reads like a check. It's clearly documented in the docstring and the alternative (a separate tracking pass) would be heavier, so this is a reasonable trade-off — just flagging for future readers.

4. **Spec delta is large relative to the code** (~90 of the added lines are `specs/021-backtest-menu-hardcoded/spec.md`). It's thorough and consistent with the prior update pattern, so this is fine — just noting the behavioral change is smaller than the diff stat suggests.

Overall this looks good to merge once the required checks pass.

---

RNiveau:
Thanks for the thorough review. Addressed the one concrete nit:

- **(2) `int` vs `Optional[float]`** — fixed in `de6887c`: `time_cut_min_favorable_points` now passes `5.0` in both the `B9HTC` definition and the test fixture, matching the annotation.

Leaving the other three as-is, per your assessment:
- **(1) 0-point time-cut → loss** — inherited from the existing `END_OF_DAY` convention and consistent with FR-031; changing it would be a broader classification decision beyond this PR.
- **(3) side-effect in `_resolve_time_cut`** — kept; the MFE update lives with the deadline check it feeds, documented in the docstring, and a separate tracking pass would be heavier.
- **(4) spec delta size** — intentional, matching the prior short-positions / parametrization amendment pattern.

Tests still green locally (78 passed); CI re-running on the new commit.

---

api/services/backtest_service.py
```diff
@@ -632,6 +668,45 @@ def _resolve_exit(
...
+        """Time-based cut for the "Bougie de 9h (time cut)" variant.
+
+        Tracks the position's max favorable excursion candle by candle.
+        Once time_cut_minutes have elapsed since entry, if the trade has
+        never moved more than time_cut_min_favorable_points in its favor,
+        it is closed at market (this candle's close). "Never been higher
+        than N points" means the cut also fires when the best move was
+        exactly N points, so the comparison is <=.
+        """
+        assert position.time_cut_min_favorable_points is not None
```
RNiveau: Don't use assert in the code. Add this in the constitution if needed

tests/api/services/test_backtest_service.py
```diff
@@ -163,13 +171,23 @@ def test_m5_fetch_raising_returns_no_trade(self):
...
         service = make_service([], [])
         definitions = service.list_definitions()
-        assert len(definitions) == 1
+        assert len(definitions) == 2
         assert definitions[0].code == "B9H"
         assert definitions[0].display_name == "CAC40 Bougie de 9h"
         assert definitions[0].instrument == "FRA40.I"
+        assert definitions[0].time_cut_minutes is None
+
+    def test_list_definitions_includes_the_time_cut_variant(self):
```
RNiveau: This test is useless

RNiveau:
Addressed both review comments:

**No `assert` in code** (`_resolve_time_cut`)
- Removed the two type-narrowing asserts, replacing them with local vars + an early-return guard (`c29e3e4`).
- Extended it project-wide per your suggestion: the 4 pre-existing `assert isinstance(report_order, ReportOrder)` uses in `report_service.py`, `binance_report_service.py`, and `saxo_order/commands/get_report.py` now raise explicit `TypeError` (`23a9009`).
- Added a **"No assert in Production Code"** rule to the constitution (Clean Code First + Backend Code Quality Standards), bumped to **1.3.0**. There are now no asserts left in production code.

**"This test is useless"** (`test_list_definitions_includes_the_time_cut_variant`)
- Removed it — it only asserted the hardcoded `B9HTC` definition matched its own constants. Time-cut behavior stays covered by `TestTimeCutVariant`.

Backend suite green locally (298 passed, 5 pre-existing skips); black/flake8 clean. CI re-running.

---

PR #659, opened 2026-07-24
feat: GER40 Bougie de 9h double take-profit backtest (spec 025)

RNiveau:
## Summary

Adds a third hardcoded backtest — **"GER40 Bougie de 9h"** — to the Backtest menu (spec 025). It reuses the entire "CAC40 Bougie de 9h" engine (9:00–10:00 Paris-local H1 reference range, both-direction 5-minute breakout/reversal detection, entry validity, gap-fill exits, one-position-at-a-time, range/day/CSV outputs) and layers on the GER40-specific rules.

GER40.I trades the same 9:00–17:30 CET session as FRA40.I, so the existing `EUMarket`/Paris-window logic applies unchanged.

## What's different from CAC40

- **Instrument** → `GER40.I`
- **Defaults** → take-profit offset **10**, break-even trigger **50**, max entry distance **40**, stop **150**
- **Double take-profit / two lots** → every entry opens two lots: the first exits at the **H1 midpoint** (TP1 = `(high+low)/2`), the runner at the **full target** (TP2 = high − 10 / low + 10); when TP1 fills, the runner's stop moves to **break-even**
- **Stop from the H1 reference level** → 150 pts below the H1 low (long) / above the H1 high (short), shared by both lots — so a stop before TP1 closes both (the "SL is x2" loss)
- **One aggregated `Trade`** per position (net points = both lots summed); the summary classifies double-TP positions by **net-points sign** (a TP1-then-break-even runner is a net-positive win)

## Changes

**Model / API**
- `Strategy.G9H` enum
- `BacktestDefinition` gains `default_parameters`, `double_take_profit`, `first_target_fraction`, `stop_from_reference_level`
- `resolve_parameters` merges per-run overrides onto each definition's own defaults; the router resolves thresholds **per-definition** (GER40 → 150/10/50/40, CAC40 → 50/10/20/20)
- `/definitions` response exposes `default_parameters` + `double_take_profit`

**Engine** (`api/services/backtest_service.py`)
- `_OpenPosition` two-lot state + reference-level stop
- `_resolve_exit_double` (TP1 partial + runner break-even/TP2/stop, +50 arming) and `_close_double_trade` (aggregated net points)
- `_build_summary` net-points-sign classification for double-TP definitions

**Frontend**
- `BacktestDefinition` interface gains the new fields; the page pre-fills threshold inputs from the selected definition's defaults. The GER40 entry, run, detail view, and CSV export all work through the existing screens (no response-shape change).

## Backward compatibility

**CAC40 B9H/B9HTC are byte-for-byte unchanged** — every new behavior is gated behind the `double_take_profit` / `stop_from_reference_level` per-definition flags. No new endpoints.

## Testing

- 15 new tests: double-TP engine (both-lots stop-out, +50 break-even, TP1-then-break-even, TP1-then-TP2 full winner, end-of-day, short mirror, multi-position day, no-trade, no-data), net-sign summary classification, and per-definition parameter resolution / CSV export at the router
- `poetry run pytest` → **473 passed, 10 skipped**
- `black` / `isort` / `flake8` / `mypy` clean; frontend `tsc + vite build` and `eslint` clean

## Spec

Full spec-driven artifact set under `specs/025-ger40-bougie-9h/` (spec, plan, research, data-model, contracts, quickstart, tasks), already on `main` via #655.

## One thing to confirm on review

The stop is placed **150 points below the H1 low** (level-based, per "SL 150 below the lower"), a deliberate difference from CAC40's entry-relative stop — flagged in the spec Clarifications.

https://claude.ai/code/session_016TJ6vDqK7ZrXkvaTePtXxx

---
_Generated by [Claude Code](https://claude.ai/code/session_016TJ6vDqK7ZrXkvaTePtXxx)_


RNiveau (commented):
## Review — GER40 Bougie de 9h double take-profit

Verified locally on `claude/ger40-backtest-spec-025-k9togf` @ `a9915d8`: `poetry run pytest` → **473 passed, 10 skipped**.

Backward compatibility for `B9H` / `B9HTC` looks genuinely airtight — every new behavior is gated on `double_take_profit` / `stop_from_reference_level`, and the single-lot path is untouched. The per-definition parameter resolution is a clean refactor.

One blocking bug, plus three smaller items.

---

### 🔴 1. TP1 can sit *behind* the entry, banking a loss as a "take profit"

`api/services/backtest_service.py:611` sets TP1 to the H1 midpoint unconditionally, but nothing in `_is_valid_long_entry` / `_is_valid_short_entry` (`:108`, `:126`) constrains the entry relative to that midpoint — only distance from the reference level (≤ 40) and being on the correct side of TP2.

So whenever the **H1 range is under 80 points** (2 × `max_entry_distance`), an entry can open *past* the midpoint. GER40's 9:00–10:00 range is frequently 50–80 pts, so this is not a corner case.

Reproduced with H1 8000–8050 (TP1 = 8025, TP2 = 8040) and a valid long entry at 8030:

```
Buy  entry 8030  exit 8020  break_even  points -12.0
```

What happens:

1. On the next candle `tp1_hit` is trivially true (price is already past the midpoint).
2. `_target_fill` fills at the candle open (8028), so the first lot's leg is banked as **−2 points** — a "take profit" that lost money.
3. `first_target_taken` → `be_armed = True` immediately (`:466`), so the runner's 150-pt reference stop is discarded on the very first candle.
4. The whole position degenerates to an instant break-even stop and closes at **−12 points** with `exit_reason=BREAK_EVEN`.

Net effect: the 150-pt structural stop — the entire point of the GER40 variant — never gets a chance to work on narrow-range days, and the summary silently books these as losses.

**Suggested fix:** reject entries already at/beyond the first target, mirroring the existing `entry_price < take_profit_level` guard; or clamp/skip the TP1 leg when it isn't favorable relative to entry.

### 🟡 2. `_stop_fill` / `_target_fill` are duplicated rather than reused

The two new helpers at `:496` and `:507` are behaviorally identical to the inline gap-fill blocks still present in `_resolve_exit` (`:737`–`:765`). Since they were extracted anyway, the single-lot path should call them — otherwise the two copies will drift.

### 🟡 3. `double_take_profit` + `time_cut_minutes` silently ignores the time cut

`_resolve_exit_double` never calls `_resolve_time_cut`, and `max_favorable_points` is never updated on the double path. Harmless today (G9H sets no time cut), but it's a silent no-op for any future definition that sets both. Worth a comment, or an assertion at definition-registration time.

### 🟡 4. `double_take_profit` is plumbed to the frontend but never used

`frontend/src/services/api.ts:962` declares it; nothing in `Backtest.tsx` reads it.

Meanwhile `_close_double_trade` deliberately returns a `Trade` where `points ≠ exit_price − entry_price` (see its docstring at `:939`) — a real trap in the day-detail table and the CSV export, where a reader sees `entry 8015 / exit 8040 / points 35` and cannot reconcile the three numbers. Either surface a "2 lots" badge or a second column for double-TP definitions, or drop the unused field.

---

### On the question raised in the description

The level-based stop (150 points below the H1 low) is implemented exactly as described and is cleanly gated behind `stop_from_reference_level`. The approach is coherent — but see finding #1: as currently written that stop is frequently discarded on the first candle after entry, so a range test over real data would not actually be measuring it.

---

RNiveau:
Thanks for the thorough review — all four addressed in `1731bbf`.

### 🔴 1. TP1 behind the entry — fixed

You're right, and it's the important one. Entry validity now requires the entry to sit on the favorable side of TP1 (the H1 midpoint) for double-TP definitions, mirroring the existing `entry < take_profit_level` guard: `entry < TP1` for a long, `entry > TP1` for a short (`_is_valid_long_entry` / `_is_valid_short_entry`, threaded through `_DirectionSearch.first_target_level`). So a narrow-range entry that would open past the midpoint is now rejected rather than firing TP1 immediately and discarding the 150-pt stop. Your exact repro (H1 8000–8050, entry 8030) now returns no trade.

This refines the spec's "50% level at or below the entry (degenerate)" edge case — which assumed fire-immediately was acceptable and rare — to *reject* the entry instead, since as you note it's frequent for GER40 and defeats the structural stop. I'll fold that back into the spec's edge-case wording.

Regression locked with `TestDoubleTakeProfitEntryValidity` (long + short rejection, and a below-midpoint control that still trades).

### 🟡 2. Duplicated gap-fill helpers — fixed

`_resolve_exit` (single-lot) now calls `_stop_fill` / `_target_fill` instead of the inline blocks, so there's one implementation. B9H/B9HTC behavior is unchanged (existing exit tests still green).

### 🟡 3. `double_take_profit` + time cut — fixed

`BacktestDefinition.__post_init__` now raises `ValueError` if a definition sets both, since the double-TP exit path doesn't evaluate the cut — fail-fast at registration rather than a silent no-op (explicit exception, not `assert`, per the constitution). Covered by `TestBacktestDefinitionValidation`.

### 🟡 4. Unused flag + points reconciliation trap — fixed

The `double_take_profit` flag is now used: the day-detail view shows a "2 lots per entry — Points is the net of both lots, so it need not equal exit − entry" note for double-TP definitions, so the `entry 8015 / exit 8040 / points 35` row reconciles for the reader.

---

Local checks against the CI commands: `black`/`isort`/`flake8`/`mypy` clean, `pytest` → **477 passed, 10 skipped**, frontend `tsc + vite build` and `eslint` clean.

---

PR #670, opened 2026-07-24
feat: add backtest candle caching with DynamoDB

RNiveau:
## Summary

Implements a caching layer for backtest candle data using DynamoDB to avoid redundant Saxo API calls when evaluating the same trading day multiple times. This addresses FR-036 through FR-040 requirements for cost optimization without affecting correctness.

## Key Changes

- **Async backtest evaluation**: Converted `BacktestService.evaluate_day()` and `run_range()` to async methods to support cache operations
  - Updated all test methods to use `async def` and `await` calls
  - Modified router endpoints to `await` the async service methods

- **DynamoDB cache integration**:
  - Added `get_cached_backtest_candles()` and `store_backtest_candles()` methods to `DynamoDBClient`
  - Implemented `_get_cached_candles()` and `_store_candles()` in `BacktestService` with graceful degradation (cache misses/failures fall back to Saxo)
  - Added `_candle_from_cached_dict()` to reconstruct `Candle` objects from cached DynamoDB items, converting Decimal prices back to float

- **New data model**: Added `CachedDayCandles` dataclass to represent cached raw candle data (H1 reference + 5-minute session candles) without computed trades

- **Infrastructure**: Created `backtest_candle_cache` DynamoDB table with `definition_code` and `trading_date` as composite key, no TTL (historical data never changes)

- **Dependency injection**: Updated `BacktestService` constructor to accept optional `DynamoDBClient` and modified `get_backtest_service()` dependency to inject it

## Implementation Details

- Cache lookup happens before Saxo fetches; on miss, candles are fetched and stored for future requests
- Strategy evaluation always runs fresh against cached or fetched candles (FR-039)
- Cache outages never break backtest runs—missing client or DynamoDB errors log warnings and fall back to Saxo
- Float-to-Decimal conversion handled automatically by existing `_convert_floats_to_decimal()` utility
- All 100+ backtest tests converted to async with proper `await` syntax

https://claude.ai/code/session_01L22im68hZx7KQG7TexS6NP

### DIFF
--- a/api/services/backtest_service.py
+++ b/api/services/backtest_service.py
@@ -1,5 +1,5 @@
 import datetime
-from typing import Any, Dict, List, Optional
+from typing import List, Optional
 from zoneinfo import ZoneInfo
 
 from client.aws_client import DynamoDBClient, DynamoDBOperationError
@@ -66,6 +66,23 @@ BACKTEST_DEFINITIONS: List[BacktestDefinition] = [
         structural_stop=True,
     ),
 ]
+
+# Bump whenever a change to a BacktestDefinition's instrument, 9h
+# reference window, or session windows would change what candles are
+# fetched under the same code (e.g. re-pointing a definition at a
+# different instrument). Without this, an edited definition would
+# otherwise keep serving cache entries fetched under its old shape
+# forever - there is no TTL on the candle cache (FR-040).
+CACHE_SCHEMA_VERSION = 1
+
+
+def _cache_key(definition: BacktestDefinition) -> str:
+    """Cache key for the raw-candle cache (FR-036), covering not just
+    the definition's code but also its instrument and the schema
+    version, so a definition edit or a strategy-shape change can
+    invalidate old entries just by bumping CACHE_SCHEMA_VERSION."""
+    return f"{definition.code}:{definition.instrument}:v{CACHE_SCHEMA_VERSION}"
+
 
 # Candle horizons for the two Saxo fetches. Unlike the strategy
 # thresholds (now tunable via BacktestParameters), these describe the
@@ -393,14 +410,16 @@ class BacktestService:
         written to the cache, so it doesn't permanently poison later
         runs over the same day."""
         params = params or BacktestParameters()
-        cached = await self._get_cached_candles(definition.code, trading_date)
+        cached = await self._get_cached_candles(
+            _cache_key(definition), trading_date
+        )
 
         if cached is not None and cached.has_data and cached.h1_candle is None:
             # _get_cached_candles already guards against this, but a
             # cache problem must never break a backtest - fall through
             # to a fresh fetch instead of failing the request.
             self.logger.warning(
-                f"Cached backtest entry for {definition.code}/"
+                f"Cached backtest entry for {_cache_key(definition)}/"
                 f"{trading_date} has has_data=True but no h1_candle; "
                 "treating as a miss"
             )
@@ -432,7 +451,7 @@ class BacktestService:
             return DayResult(date=trading_date, status=DayStatus.NO_DATA)
         if fetched_h1_candle is None:
             await self._store_candles(
-                definition.code, trading_date, has_data=False
+                _cache_key(definition), trading_date, has_data=False
             )
             return DayResult(date=trading_date, status=DayStatus.NO_DATA)
 
@@ -443,7 +462,7 @@ class BacktestService:
             definition, fetched_h1_candle.higher, fetched_h1_candle.lower
         ):
             await self._store_candles(
-                definition.code,
+                _cache_key(definition),
                 trading_date,
                 has_data=True,
                 h1_candle=fetched_h1_candle,
@@ -466,7 +485,7 @@ class BacktestService:
             five_min_candles = []
         else:
             await self._store_candles(
-                definition.code,
+                _cache_key(definition),
                 trading_date,
                 has_data=True,
                 h1_candle=fetched_h1_candle,
@@ -474,7 +493,11 @@ class BacktestService:
             )
 
         return self._evaluate_from_candles(
-            definition, trading_date, params, fetched_h1_candle, five_min_candles
+            definition,
+            trading_date,
+            params,
+            fetched_h1_candle,
+            five_min_candles,
         )
 
     @staticmethod
@@ -1059,10 +1082,10 @@ class BacktestService:
         )
 
     async def _get_cached_candles(
-        self, definition_code: str, trading_date: datetime.date
+        self, cache_key: str, trading_date: datetime.date
     ) -> Optional[CachedDayCandles]:
-        """Look up the raw-candle cache for (definition_code,
-        trading_date) (FR-036). Returns None on a cache miss, or when no
+        """Look up the raw-candle cache for (cache_key, trading_date)
+        (FR-036). Returns None on a cache miss, or when no
         DynamoDBClient is available (local/test usage, or a DynamoDB
         failure) - in every such case the caller falls back to fetching
         from Saxo, so a cache outage never breaks a backtest run."""
@@ -1070,48 +1093,57 @@ class BacktestService:
             return None
         try:
             item = await self.dynamodb_client.get_cached_backtest_candles(
-                definition_code, trading_date.isoformat()
+                cache_key, trading_date.isoformat()
             )
         except DynamoDBOperationError as e:
             self.logger.warning(
                 f"Backtest candle cache lookup failed for "
-                f"{definition_code}/{trading_date}: {e}"
+                f"{cache_key}/{trading_date}: {e}"
             )
             return None
         if item is None:
             return None
 
-        has_data = bool(item["has_data"])
-        if not has_data:
-            return CachedDayCandles(has_data=False)
-        return CachedDayCandles(
-            has_data=True,
-            h1_candle=self._candle_from_cached_dict(item["h1_candle"]),
-            m5_candles=[
-                self._candle_from_cached_dict(c)
-                for c in item.get("m5_candles", [])
-            ],
-        )
+        try:
+            has_data = bool(item["has_data"])
+            if not has_data:
+                return CachedDayCandles(has_data=False)
+            return CachedDayCandles(
+                has_data=True,
+                h1_candle=Candle.from_dict(item["h1_candle"]),
+                m5_candles=[
+                    Candle.from_dict(c) for c in item.get("m5_candles", [])
+                ],
+            )
+        except (KeyError, ValueError, TypeError) as e:
+            # A cache problem must never break a backtest: an item
+            # written under an earlier schema (or otherwise malformed)
+            # is treated as a miss, the same as if nothing were cached.
+            self.logger.warning(
+                f"Malformed backtest cache item for "
+                f"{cache_key}/{trading_date}: {e}"
+            )
+            return None
 
     async def _store_candles(
         self,
-        definition_code: str,
+        cache_key: str,
         trading_date: datetime.date,
         has_data: bool,
         h1_candle: Optional[Candle] = None,
         m5_candles: Optional[List[Candle]] = None,
     ) -> None:
-        """Store the raw candles fetched for (definition_code,
-        trading_date) so a later request for the same pair is served
-        from the cache (FR-037/FR-038). A missing DynamoDBClient or a
-        DynamoDB failure degrades to "not cached this time" rather than
-        failing the backtest - caching is a cost optimization, not a
-        correctness requirement."""
+        """Store the raw candles fetched for (cache_key, trading_date)
+        so a later request for the same pair is served from the cache
+        (FR-037/FR-038). A missing DynamoDBClient or a DynamoDB failure
+        degrades to "not cached this time" rather than failing the
+        backtest - caching is a cost optimization, not a correctness
+        requirement."""
         if self.dynamodb_client is None:
             return
         try:
             await self.dynamodb_client.store_backtest_candles(
-                definition_code,
+                cache_key,
                 trading_date.isoformat(),
                 has_data,
                 h1_candle.to_dict() if h1_candle is not None else None,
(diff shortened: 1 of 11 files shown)


api/services/backtest_service.py
```diff
@@ -354,27 +361,58 @@ def get_definition(self, code: str) -> Optional[BacktestDefinition]:
...
+        else:
+            h1_start_utc, h1_end_utc = paris_reference_window_utc(trading_date)
+            fetched_h1_candle = self._fetch_h1_reference_candle(
+                definition.instrument, h1_start_utc, h1_end_utc
+            )
+            if fetched_h1_candle is None:
+                await self._store_candles(
+                    definition.code, trading_date, has_data=False
+                )
+                return DayResult(date=trading_date, status=DayStatus.NO_DATA)
```
RNiveau: **Blocking: a transient Saxo failure gets cached permanently as NO_DATA.**

`_fetch_h1_reference_candle` returns `None` for two very different things: Saxo genuinely has no candle for that day, *and* a caught `SaxoException` — expired token, rate limit, network blip. There's even a test for the second case (`test_h1_fetch_raising_returns_no_data`).

This code writes `has_data=False` for both. Since the table has no TTL and there is no invalidation path, one blip in the middle of a multi-year `run_range` permanently marks real trading days as NO_DATA, and every later backtest over that range silently reads a distorted result with no way to notice.

Suggest making the fetch distinguish "no data" from "fetch failed" (re-raise, or return a sentinel), and only writing the marker for the former.

api/services/backtest_service.py
```diff
@@ -354,27 +361,58 @@ def get_definition(self, code: str) -> Optional[BacktestDefinition]:
...
+            five_min_candles = self._fetch_five_minute_candles(
+                definition.instrument, h1_end_utc, session_end_utc
+            )
+            await self._store_candles(
+                definition.code,
+                trading_date,
+                has_data=True,
+                h1_candle=h1_candle,
+                m5_candles=five_min_candles,
+            )
```
RNiveau, line 493: **Blocking: same failure-caching problem on the 5-minute fetch.**

`_fetch_five_minute_candles` swallows `SaxoException` and returns `[]`. Before this PR that made the day a *transient* NO_TRADE, recoverable on the next run. Now it is persisted as `has_data=True, m5_candles=[]`, so the day is permanently NO_TRADE with a valid-looking H1 range — the existing test `test_m5_fetch_raising_returns_no_trade` documents exactly this path.

Skip the `_store_candles` call when the m5 fetch failed.

api/services/backtest_service.py
```diff
@@ -354,27 +361,58 @@ def get_definition(self, code: str) -> Optional[BacktestDefinition]:
...
-            return DayResult(date=trading_date, status=DayStatus.NO_DATA)
+        cached = await self._get_cached_candles(definition.code, trading_date)
+        if cached is not None:
+            if not cached.has_data:
+                return DayResult(date=trading_date, status=DayStatus.NO_DATA)
+            if cached.h1_candle is None:
+                raise SaxoException(
+                    "Cached backtest entry has has_data=True but no "
+                    "h1_candle"
+                )
```
RNiveau: This contradicts the stated contract that a cache problem never breaks a backtest — a malformed cached item takes down the whole run instead of degrading to a Saxo fetch. Treat it as a cache miss and fall through to the `else` branch.

Same class of issue at `_get_cached_candles` below (`item["has_data"]` / `item["h1_candle"]`): an item written by an earlier schema raises `KeyError`, which isn't caught by the `except DynamoDBOperationError`. Any item you can't read should be a miss.

pulumi/dynamodb.py
```diff
@@ -105,3 +105,21 @@ def alert_digests_table() -> aws.dynamodb.Table:
...
+
+def backtest_candle_cache_table() -> aws.dynamodb.Table:
+    return aws.dynamodb.Table(
+        "backtest_candle_cache",
+        attributes=[
+            aws.dynamodb.TableAttributeArgs(name="definition_code", type="S"),
+            aws.dynamodb.TableAttributeArgs(name="trading_date", type="S"),
+        ],
+        hash_key="definition_code",
+        range_key="trading_date",
```
RNiveau, line 118: The cache key is `(definition_code, trading_date)`, but the cached bytes actually depend on the definition's `instrument` and its 9h/session windows. If any of those is ever edited while keeping the same code, the cache silently serves candles fetched under the old config — and with no TTL, forever.

Worth folding an instrument or `cache_version` component into the hash key so a definition change naturally invalidates.

pulumi/dynamodb.py
```diff
@@ -105,3 +105,21 @@ def alert_digests_table() -> aws.dynamodb.Table:
...
+        name="backtest_candle_cache",
+        billing_mode="PAY_PER_REQUEST",
+        stream_enabled=True,
```
RNiveau: Nit: streams look copy-pasted from `alert_digests_table`. Nothing consumes this stream, and it's a paid feature on a table whose entire purpose is cost reduction. Suggest dropping `stream_enabled` / `stream_view_type`.

api/services/backtest_service.py
```diff
@@ -855,3 +893,91 @@ def _close_trade(
...
+                ),
+            )
+        except DynamoDBOperationError as e:
+            self.logger.warning(
+                f"Backtest candle cache store failed for "
+                f"{definition_code}/{trading_date}: {e}"
+            )
+
+    @staticmethod
+    def _candle_from_cached_dict(data: Dict[str, Any]) -> Candle:
```
RNiveau: Nit: this duplicates the existing `Candle.from_dict` (`model/workflow.py:205`) — the only difference is the `float()` coercion of the Decimals DynamoDB hands back. Adding `float(...)` in `Candle.from_dict` and deleting this private copy keeps one serialization round-trip for `Candle`.

RNiveau (commented):
Reviewed on the branch: full suite passes (491 passed, 10 skipped), flake8 clean, mypy clean apart from the pre-existing `aioboto3` stub warning.

The design is sound — caching *raw candles only*, never a computed `DayResult`, is the right call, and the DI / graceful-degradation shape matches the rest of the codebase. I also checked that a partial in-progress session can't be poisoned into the cache: `_parse_date` already rejects an unclosed today for both the day and range endpoints. Cache tests are good.

Two blocking issues, both about **what gets written on failure**. With no TTL and no invalidation path, a transient Saxo error (expired token, rate limit, network blip) is persisted as a permanent wrong answer:

1. An H1 fetch that *raised* is stored as `has_data=False` → the day is permanently NO_DATA.
2. An m5 fetch that *raised* is stored as `has_data=True, m5_candles=[]` → the day is permanently NO_TRADE.

Both silently distort every later backtest over that range. Details inline, plus three smaller points (cache key doesn't cover the definition's instrument/windows, a `SaxoException`/`KeyError` path that breaks the "cache never breaks a backtest" contract, and a duplicate of `Candle.from_dict`).

Heads-up on merge: this is based on `main`, which doesn't yet have 63640e6 (wide-range `min_h1_range_points`). That commit returns NO_TRADE *before* the m5 fetch, in the region this PR restructures into an if/else — make sure the min-range check still runs on the cache-hit path.

api/services/backtest_service.py
RNiveau: Fixed in d581916: `_fetch_h1_reference_candle` no longer swallows `SaxoException` — it now propagates to `evaluate_day`, which returns NO_DATA for that request but skips `_store_candles` entirely, so a transient failure isn't persisted. Only a genuine empty result from Saxo (no candle for the window) is cached as `has_data=False`. Added `test_h1_fetch_failure_is_not_cached` to cover this.

---

api/services/backtest_service.py
RNiveau: Fixed in d581916: `_fetch_five_minute_candles` now lets `SaxoException` propagate; `evaluate_day` catches it, falls back to an empty candle list for that request, and skips the `_store_candles` call — only a successful (even if empty) fetch is cached. Added `test_m5_fetch_failure_is_not_cached` to cover this.

---

api/services/backtest_service.py
RNiveau: Fixed in d581916: `_get_cached_candles` now wraps the item reconstruction in `except (KeyError, ValueError, TypeError)`, treating any unreadable/malformed cached item as a miss (falls through to a fresh Saxo fetch) instead of raising. The `has_data=True`/`h1_candle is None` defensive branch in `evaluate_day` now falls back to a fresh fetch too, rather than raising, as a second line of defense. Added `test_malformed_cache_item_falls_back_to_saxo`.

---

pulumi/dynamodb.py
RNiveau: Fixed in d581916: added a module-level `CACHE_SCHEMA_VERSION` and a `_cache_key(definition)` helper that folds `code`, `instrument`, and `v{CACHE_SCHEMA_VERSION}` into the string stored under the table's `definition_code` attribute. Editing a definition's instrument (or bumping the version when the strategy/window shape changes) now naturally orphans old entries instead of silently serving them forever. Added `test_cache_key_covers_code_instrument_and_version`.

---

pulumi/dynamodb.py
RNiveau: Fixed in d581916: dropped `stream_enabled`/`stream_view_type` from `backtest_candle_cache_table`.

---

api/services/backtest_service.py
RNiveau: Fixed in d581916: added the `float(...)` coercion to `Candle.from_dict` (model/workflow.py) and deleted the private `_candle_from_cached_dict` copy — `_get_cached_candles` now calls `Candle.from_dict` directly.

---

RNiveau:
Re: your comment "shouldn't be optional" on `api/services/backtest_service.py` (https://github.com/RNiveau/saxo-order/pull/670#discussion_r3648086910) — the inline reply anchor became invalid after I rebased onto main, so replying here instead.

Could you clarify which parameter this refers to? A couple of candidates near that line:
- `BacktestService.__init__`'s `dynamodb_client: Optional[DynamoDBClient] = None` — made optional deliberately so local/dev/test usage without AWS still works (caching is a best-effort optimization, not a hard requirement).
- `CachedDayCandles.h1_candle: Optional[Candle]` (model/backtest.py) — `Optional` there because a `has_data=False` entry legitimately has no candle.

Happy to change whichever you meant once I know which one — didn't want to guess and touch the wrong thing.

All six other findings are addressed in d581916 (replied inline on each thread): transient Saxo failures on both the H1 and 5-minute fetch are no longer cached, a malformed/unreadable cached item now falls back to a fresh Saxo fetch instead of raising, the cache key now folds in the instrument and a bumpable schema version, the unused DynamoDB stream is dropped, and the private candle-deserialization helper is gone in favor of `Candle.from_dict`. Also rebased onto main's wide-range structural-stop variant so the H1-range filter runs (and skips the m5 fetch) on both the cache-hit and cache-miss paths. Full suite (504 passed, 10 skipped), flake8, and mypy are clean.

---

tests/client/test_aws_client_backtest_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        mock_table.get_item.return_value = {
+            "ResponseMetadata": {"HTTPStatusCode": 200},
+            "Item": {
+                "definition_code": "B9H",
+                "trading_date": "2026-07-14",
+                "has_data": True,
+            },
+        }
+
+        item = await client.get_cached_backtest_candles("B9H", "2026-07-14")
```
RNiveau: you're testing a mock

tests/client/test_aws_client_backtest_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        assert item is not None
+        assert item["has_data"] is True
+
+    async def test_returns_none_on_miss(self, mock_dynamodb_resource, client):
+        _, mock_table = mock_dynamodb_resource
+        mock_table.get_item.return_value = {
+            "ResponseMetadata": {"HTTPStatusCode": 200},
+        }
+
+        item = await client.get_cached_backtest_candles("B9H", "2026-07-14")
```
RNiveau: same

tests/client/test_aws_client_backtest_cache.py
```diff
@@ -0,0 +1,127 @@
...
+        assert item is None
+
+    async def test_returns_none_on_error_response(
+        self, mock_dynamodb_resource, client
+    ):
+        _, mock_table = mock_dynamodb_resource
+        mock_table.get_item.return_value = {
+            "ResponseMetadata": {"HTTPStatusCode": 500},
+        }
+
```
RNiveau: same

api/services/backtest_service.py
RNiveau: Fair — fixed in 4da4b87: removed `TestGetCachedBacktestCandles` entirely. Those three tests just echoed back a canned mock response (or None on a configured miss/error), unlike `TestStoreBacktestCandles` which verifies real transformation logic (the float→Decimal conversion). The status-check/`.get("Item")` shape is the same generic pattern every other read method on `DynamoDBClient` uses, and `_dynamo_operation`'s error wrapping is already covered generically in `test_aws_client.py`.

---

tests/client/test_aws_client_backtest_cache.py
RNiveau: Fair — fixed in 4da4b87: removed `TestGetCachedBacktestCandles` entirely. Those three tests just echoed back a canned mock response (or None on a configured miss/error), unlike `TestStoreBacktestCandles` which verifies real transformation logic (the float→Decimal conversion). The status-check/`.get("Item")` shape is the same generic pattern every other read method on `DynamoDBClient` uses, and `_dynamo_operation`'s error wrapping is already covered generically in `test_aws_client.py`.

---

tests/client/test_aws_client_backtest_cache.py
RNiveau: Same fix — see the reply on the sibling thread: `TestGetCachedBacktestCandles` (all three tests) removed in 4da4b87.

---

tests/client/test_aws_client_backtest_cache.py
RNiveau: Same fix — see the reply on the sibling thread: `TestGetCachedBacktestCandles` (all three tests) removed in 4da4b87.

---

api/services/backtest_service.py
```diff
@@ -355,9 +374,14 @@ def _feed_short(self, candle: Candle) -> Optional[float]:
 class BacktestService:
     """Runs the hardcoded backtests exposed by the Backtest menu."""
 
-    def __init__(self, candles_service: CandlesService):
+    def __init__(
+        self,
+        candles_service: CandlesService,
+        dynamodb_client: Optional[DynamoDBClient] = None,
```
RNiveau: it shouldn't be optional

RNiveau: Fixed in 0cb9f5d: `dynamodb_client` is now a required `DynamoDBClient` param (no `Optional`, no default), matching `AssetDetailsService`/`WorkflowService`.

To keep local/dev usage without AWS working (the reason it was `Optional` in the first place), added `get_dynamodb_client_best_effort` in `api/dependencies.py`: unlike `get_dynamodb_client` it doesn't raise 403 outside an AWS context, and unlike `get_dynamodb_client_optional` it's never `None` — it always constructs a real `DynamoDBClient`, whose underlying resource may simply be unset. A call through it then raises `RuntimeError`, which `_get_cached_candles`/`_store_candles` now catch alongside `DynamoDBOperationError`, degrading to the same cache miss/no-op as before. Tests updated to construct a `DynamoDBClient(dynamodb_resource=None)` (`NO_CACHE_CLIENT`) wherever a test doesn't care about caching, instead of passing `None`.

---

PR #672, opened 2026-07-26
refactor: split the backtest service into a composable package

RNiveau:
Breaks the 1400-line `BacktestService` god-class into `api/services/backtest/`, guarded by a golden characterization test added **before** any code moved.

Four backtests had accumulated six variant flags, three hand-written exit resolvers and long/short logic written out twice — adding a fifth meant threading another boolean through all of it.

## Commits

| Commit | What |
|---|---|
| `5bec9f8` | Golden snapshot: 4 definitions × 2-month synthetic market → summary, day rows, exit-reason histogram, one detail day's trades |
| `3c1f90f` | Extract non-orchestrating statics → `definitions` / `calendar` / `candles` / `analytics` / `statistics` |
| `f986c23` | Extract all I/O into `CandleSource` (`evaluate_day`: ~110 lines → 6) |
| `69e21d4` | Collapse long/short duplication onto `Side.favorable()`; new `side` / `entry` / `position` |
| `a2c8a1d` | Replace the 3 exit resolvers with an ordered policy chain built in `rules.py` |
| `829becf` | `SingleLot` / `TwoLot` own points arithmetic + win/loss classification; `Position` 14 → 8 ctor args |
| `0a4662c` | **Behavior fix** — default-parameter fallback + `BacktestDefinition` validation |
| `3c38843` | Review follow-ups |

## The one behavior change

`0a4662c` is the only commit that changes behavior, in two ways:

- `evaluate_day` / `run_range` called without explicit parameters fell back to the `BacktestParameters` **class** defaults, not the definition's. The router always pre-resolves, so this never fired in production — but a direct call for GER40 would have run on CAC40's 50/10/20/20 instead of 150/10/50/40.
- `BacktestDefinition.__post_init__` now rejects every flag combination it cannot honor in full, not just the one: double take-profit with no first target, a first-target fraction without double take-profit, a fraction outside the H1 range, and a half-configured time cut.

Everything else is behavior-preserving.

## Bugs removed by construction

Exit routing was `if double: … elif structural and not armed: …`, so a definition setting two variant flags silently lost one:

- `structural_stop` + `double_take_profit` → the structural stop was dropped
- `structural_stop` + a time cut → the cut was dropped

No shipped definition used either combination, so no results change. Flags now compose, and `test_rules.py` pins both the shipped chains and the previously-broken pairings.

## Verification

- Snapshot generated once at `5bec9f8` and **never regenerated** — `git log --follow` on the JSON shows a single commit — and the golden test passes at every intermediate commit.
- 607 passed, 10 skipped; `flake8` / `black` / `isort` clean; `mypy` clean apart from a pre-existing `aioboto3` stub error in `api/main.py`.

## Structure

`service.py` is now orchestration only; `candle_source.py` is the only module doing I/O; `rules.py` is the only module that reads a definition's variant flags. The 2024-line test file became 17 files averaging ~140 lines, mirroring the modules.


RNiveau (commented):
## Overview

Reviewed all 7 commits individually against the pre-PR `api/services/backtest_service.py`, and verified each commit independently (tests + golden snapshot pass at every one).

Breaks the 1400-line `BacktestService` god-class into a 14-module `api/services/backtest/` package, guarded by a golden characterization test added *first*. Sequence is textbook:

| Commit | What |
|---|---|
| `5bec9f8` | Golden snapshot: 4 definitions × 2-month synthetic market → summary, day rows, exit-reason histogram, one detail day's trades |
| `3c1f90f` | Extract non-orchestrating statics → `definitions`/`calendar`/`candles`/`analytics`/`statistics` |
| `f986c23` | Extract all I/O into `CandleSource` (`evaluate_day`: ~110 lines → 6) |
| `69e21d4` | Collapse long/short duplication onto `Side.favorable()`; new `side`/`entry`/`position` |
| `a2c8a1d` | Replace 3 exit resolvers with an ordered policy chain built in `rules.py` |
| `829becf` | `SingleLot`/`TwoLot` own points arithmetic + win/loss classification; `Position` 14 → 8 ctor args |
| `0a4662c` | Fix: default-parameter fallback + `BacktestDefinition.__post_init__` validation |

## Verification

- `poetry run pytest tests` → **607 passed, 10 skipped**; `mypy` clean (4 pre-existing `import-untyped` errors in unrelated files); `flake8`/`black`/`isort` clean. CI green on both jobs.
- Golden snapshot JSON is **byte-untouched after `5bec9f8`**, and the golden test passes at every intermediate commit (161→243 tests) — the "behavior unchanged" claim holds, not just asserted.
- Hand-checked the `Side` algebra against the old paired branches: `reached`/`receded`/`closed_beyond`/`stop_fill`/`target_fill`/`worse`, `stop_level` (`entry - sign*sl`), `break_even_arm_level`, `is_valid_entry`'s three guards, and `DirectionSearch.feed`'s breach/candidate/confirm transitions all reduce to the original long and short forms exactly.
- Chain ordering reproduces the old implicit routing for all four shipped definitions, including the structural variant's armed/unarmed split (`Stop(only_when_armed=True)` before `Target`, `StructuralStop` after) and the double path's stop-before-TP1.
- No test coverage lost: 4 old test names absent, all with renamed equivalents in `test_definitions.py`.

The `a2c8a1d` commit message's claim checks out: `structural_stop + time_cut` previously dropped the cut and `structural_stop + double_take_profit` dropped the structural stop, because routing was `if double: … elif structural and not armed: …`. Nothing shipped those combos, so no results change — but the class of bug is gone, and `test_rules.py` pins it.

## Issues and suggestions

**1. `_take_profit_level` calls `reference_level` with swapped arguments** (`api/services/backtest/service.py:312`)

```python
far_level = side.reference_level(h1_low, h1_high)
```

Correct (long → `h1_high`), but it works by lying to a named-parameter API, which is exactly the kind of thing `Side` exists to eliminate. Add `Side.far_level(h1_high, h1_low)` and call it honestly — it's the one place in the refactor where argument positions had to be traced to confirm correctness.

**2. `build_lot_model`'s silent degrade is now dead code that contradicts the new validation** (`api/services/backtest/rules.py:70`)

`__post_init__` rejects `double_take_profit` without `first_target_fraction`, so the fallback is unreachable — and `tests/api/services/backtest/test_lots.py:99` can only reach it by mutating a definition post-construction. Since the PR's stated principle is "a flag that cannot take effect must not ship as a no-op", pick one: drop the fallback, or make it raise. Keeping a silent degrade *and* a constructor rejection for the same condition means two answers to one question.

**3. `StructuralStop` still silently no-ops when H1 levels are missing** (`api/services/backtest/position.py:58`, `api/services/backtest/policies.py:106`)

`Position.h1_high/h1_low` are `Optional` with `None` defaults, so `structural_level is None` makes the policy do nothing. The service always passes them, so no live bug — but it's the same anti-pattern `0a4662c` set out to remove, one layer down. Making them required on `Position` (or letting `StructuralStop` carry the level) removes the branch.

**4. The both-searches-must-be-fed invariant is undocumented** (`api/services/backtest/service.py:256`)

The comment explains the long-wins tiebreak but not why the feed loop and the selection loop are separate: `feed()` mutates search state, so both sides must advance on *every* candle even when one is about to open. A future "simplify this into one loop with an early `break`" would silently change the short side's state machine. One sentence there would prevent it.

**5. Nits**

- `_reset(searches)` (`service.py:298`) is unannotated — should be `Dict[Side, DirectionSearch]`; the rest of the package is fully typed.
- `if not self.double_take_profit and self.first_target_fraction:` (`model/backtest.py:82`) uses truthiness where the neighbours use `is None`; `first_target_fraction=0.0` falls through to the range check and reports the wrong error.
- `test_backtest_golden.py` recomputes `_snapshot()` per parametrized case (~20 full range runs for 4 assertions). Fine at 4s, but a module-scoped fixture is one line.
- `calendar.py` shadows a stdlib module name inside the package. Harmless under absolute imports; worth knowing.

**6. Housekeeping, not code**

- The PR title/body describe only the first commit. A squash-merge would land 7200 lines of refactor + a behavior fix under `test: add golden characterization tests` — retitle to the refactor, and keep the `0a4662c` fix visible in the body (it's the only commit that changes behavior).
- `specs/025-ger40-bougie-9h/data-model.md:73` still documents a `double` bool on the position, now replaced by the lot model. If specs are living docs here, that line and the new definition-validation contract need a note; if they're historical records, ignore this.

**Verdict:** good to merge after #1–#3. The safety net is the right one, it was built first, it was never regenerated, and it passes at every commit — which is what makes a refactor this large reviewable at all.

---

RNiveau:
Thanks — all of #1–#5 addressed in `3c38843`, plus the housekeeping. The golden snapshot is still byte-identical to the one committed at `5bec9f8`, so none of this changed results.

**#1 — swapped `reference_level` arguments.** Agreed, and it was the right thing to catch: correct only if you trace positions through an API whose whole purpose is to stop you doing that. Added `Side.far_level(h1_high, h1_low)` — "the end of the range this side is heading towards" — and `_take_profit_level` now calls it in argument order.

**#2 — `build_lot_model`'s silent degrade.** Made it raise. Two answers to one question is exactly the pattern the PR claims to remove, so keeping a quiet fallback behind a constructor rejection undercut the argument. `test_lots.py` now asserts the raise (still reachable only by mutating a definition post-construction, which the docstring says).

**#3 — Optional H1 levels on `Position`.** Made `h1_high`/`h1_low` required. `structural_level` returns `float` rather than `Optional[float]`, and `StructuralStop` lost its `None` branch — so the policy can no longer quietly do nothing.

**#4 — the feed-loop invariant.** Added a comment saying it outright: `feed()` advances each side's breach/candidate state machine, so merging the two loops with an early `break` would leave the short side unfed on any candle the long side opens on, changing what it sees next. That was the most useful item in the review — it's the kind of thing a future "simplify this" would walk straight into.

**#5 — nits.** `_reset` annotated `Dict[Side, DirectionSearch]`. The truthiness check is now `is not None`; you were right that `first_target_fraction=0.0` fell through to the range check and reported the wrong error. Golden snapshot is computed once per session instead of ~20 times.

**Not taken:** the `calendar.py` stdlib shadowing. The package uses absolute imports throughout so there's no live ambiguity, and renaming touches every importer for a naming preference — happy to rename it to `trading_calendar.py` if you'd rather not carry the footgun.

**#6 — housekeeping.** Retitled the PR and rewrote the body around the seven commits, with `0a4662c` called out separately as the only behavior change (both parts of it). On the specs: `2c0e4a4` amended spec 021 after its feature merged, so I read them as living and added an implementation note at the top of `specs/025-ger40-bougie-9h/data-model.md` — a table mapping the old engine names (`_OpenPosition`, `_close_double_trade`, `_DirectionSearch`) to the new modules, plus the new validation contract. I left `plan.md` / `research.md` / `tasks.md` / `quickstart.md` alone and said so in the note: those read as records of how the feature was built, not current-state docs. Say the word if you'd rather they were updated too, or if the note should live somewhere else.

---

RNiveau:
Re-reviewed `3c38843`. All six points from the review above are closed:

| Point | Fix |
|---|---|
| #1 swapped args | New `Side.far_level(h1_high, h1_low)`; `_take_profit_level` calls it honestly (`side.py:43`, `service.py:319`) |
| #2 dead silent degrade | `build_lot_model` raises `ValueError` instead of returning `SINGLE_LOT` (`rules.py:67`); the test flipped to `pytest.raises` |
| #3 `StructuralStop` no-op | `Position.h1_high/h1_low` are required floats, `structural_level` returns `float`, and the `level is None` branch is gone |
| #4 undocumented invariant | Comment at `service.py:250` spelling out that collapsing the two loops would leave the short side unfed |
| #5 nits | `_reset(searches: Dict[Side, DirectionSearch])`; `first_target_fraction is not None`; golden snapshot memoized once per session |
| #6 housekeeping | PR retitled and body rewritten with the behavior fix called out; spec 025 `data-model.md` gets an implementation note mapping the old engine names to the new ones |

Verified locally on `3c38843`:

- **607 passed, 10 skipped**; `flake8` / `black` clean; `mypy` shows the same 4 pre-existing `import-untyped` errors and nothing new.
- **The golden JSON is still untouched since `5bec9f8`** — `git log 5bec9f8..3c38843 -- tests/api/services/files/backtest_golden.json` returns zero commits, so the behavior-preserving claim survives the follow-up commit as well.

Two non-blocking observations:

- `_SNAPSHOT_CACHE` is a module-level dict rather than a `@pytest.fixture(scope="module")`. It works and the data is only ever read, but it's the one place in the suite where test state lives outside a fixture.
- Leaving `calendar.py` as-is is the right call — harmless under absolute imports, and renaming would churn the whole package.

Nothing left that I'd hold the merge for.

---

RNiveau:
Took the `_SNAPSHOT_CACHE` observation — `445d226`. It's now a `pytest_asyncio.fixture(scope="module", loop_scope="module")` instead of a memoized module-level dict, so no test state lives outside a fixture. Confirmed it still builds once rather than per case: the four parametrized cases run in 1.12s against 1.11s for a single case.

`_regenerate()` calls `_build_snapshot()` directly now, so the CLI path doesn't depend on fixture machinery.

607 passed / 10 skipped, lint clean, and the golden JSON is still untouched since `5bec9f8`.

Agreed on leaving `calendar.py` alone.

---

PR #675, opened 2026-07-27
feat: add GER40 impulsive-candle backtest (G9HIC, spec 025)

RNiveau:
A fourth GER40 variant that replaces the fixed 150-point stop with an
impulse stop: the position is never closed by price merely trading
through a level. It gives up only on a 5-minute candle that is at least
70 points wide, closes within 25% of its range from the extreme adverse
to it, and closes beyond the H1 reference level - or on its take-profit,
its armed break-even stop, or end of day. Days whose 9:00-10:00 H1 range
is not strictly wider than 70 points are not traded.

The rule is one new Side predicate plus one ExitPolicy composed in
build_exit_chain, so no engine code branches on which backtest is
running. The amplitude test alone would fire on long-wick reversal
candles that closed back inside the range, which is indecision rather
than impulse - the close-fraction test is what makes the name honest.

Also introduces EuCfdMarket (09:00-22:00 Paris) and gives every
BacktestDefinition the market its session derives from, defaulting to
EUMarket so the existing definitions keep the 9:00-17:30 cash window.
Market/USMarket/EUMarket move to model/market.py because model/__init__
imports model.backtest before defining them, which would otherwise make
BacktestDefinition.market a circular import. Two latent crashes fell out
of that: both session-end calculations passed end_minute as a minute
field, so any market with the documented end_minute > 59 (USMarket's
16:00, EuCfdMarket's 22:00) would have raised ValueError.

The golden snapshot regenerates with 569 insertions and no deletions -
every existing definition is byte-for-byte unchanged.

Claude-Session: https://claude.ai/code/session_01DEcj6YEihziuZtkT6fPsP7


tests/api/services/backtest/test_definitions.py
```diff
@@ -5,14 +5,26 @@
...
+from model import (
+    BacktestDefinition,
+    BacktestParameters,
+    EuCfdMarket,
+    EUMarket,
+)
 
 
 class TestRegistry:
     def test_lists_the_hardcoded_backtests_in_order(self):
```
RNiveau: This test is useless in the end, it tests an internal hardcoded list

RNiveau: Agreed — removed in a4424d4. It could only ever fail by being out of date with the list it mirrored, never by catching a defect, and it made adding a definition a two-file edit for no signal.

Took the two neighbours with the same smell along with it:

- `test_every_other_definition_keeps_the_cash_session` carried its own copy of the codes; it now derives the expectation from `list_definitions()` (`EuCfdMarket` for `G9HIC`, `EUMarket` for everything else), so it asserts real behaviour instead of restating the registry.
- `test_the_shipped_definitions_are_all_valid` no longer pins `len(...) == 6` — the point of that test is that importing the module builds every definition through `__post_init__`, and the count was incidental.

The per-definition tests (`test_g9hic_is_the_impulsive_candle_variant` and friends) stay: those assert thresholds the engine actually reads, and the golden snapshot is what guards the shipped behaviour end to end.

---

api/services/backtest/service.py
```diff
@@ -202,11 +201,16 @@ def _fetch_daily_candles(
         )
         try:
             return self.candles_service.build_candles(
-                instrument, UnitTime.D, EUMarket(), count, reference
+                definition.instrument,
```
RNiveau, line 217: This is the change with the widest blast radius in the PR, and I'm not sure it's the one you intended.

`_fetch_daily_candles` feeds `mm50_slope`, `adx14` and `overnight_gap` — the regime columns on every day row. Switching it from `EUMarket()` to `definition.market` means G9HIC's daily candles are now built over the 13-hour CFD window (`num_h1 = 21 - 9 + 1 = 13` in `build_daily_candles_from_h1`) instead of the 9-hour cash window every other definition uses. Three consequences:

1. **The comparison the variant exists for stops being apples-to-apples.** G9HIC vs G9HSL on the same instrument and the same dates will show different MA50 slope and different ADX, because the underlying daily bars aren't the same bars. If you segment a G9HIC export by ADX regime and compare it to a G9HSL export segmented the same way, the buckets don't line up.
2. **`overnight_gap` changes meaning.** It's `9h open - prior daily close`; the prior close is now the 22:00 CFD close rather than the 17:30 cash close, so gaps will systematically shrink for this definition only. The column header stays the same.
3. **`build_daily_candles_from_h1` gets more fragile.** It requires 13 consecutive H1 bars ending on the 21:00-local bar, and when one is missing it doesn't range-check — `_internal_build_candle(candles, i, num_h1 - 1, ...)` just takes the next 12 entries in the list, so a gap silently splices two days into one daily candle. That's pre-existing, but evening CFD bars are thinner than cash-session bars and the window is 44% longer, so it's newly likely to bite here.

The session window genuinely needs to follow the definition for the 9h window / session end / "has today closed" (that part is right). But the daily regime series is a *measurement*, not part of the strategy, and there's an argument it should stay pinned to `EUMarket()` precisely so it stays comparable across definitions. Was the switch deliberate, or collateral from threading `definition` through?

---

tests/api/services/files/backtest_golden.json
RNiveau, line 1: The golden snapshot grows by 569 additive lines, which reads as "G9HIC is now covered end to end" — but the fixture it runs against is market-agnostic, so those rows were generated *as if G9HIC ran EUMarket*.

In `tests/api/services/backtest/market_fixture.py`:

- `golden_candles_service.get_candles_in_window(code, ut, horizon, start, end)` ignores `end` entirely and returns `m5_session_candles(code, trading_date)`, which is always `SESSION_STEPS = 90` bars from `paris_reference_window_utc(d)`'s end — i.e. 10:00 → 17:25, the cash session.
- `build_candles(code, ut, market, count, reference)` ignores `market` too, so the daily-candle difference from the comment above is invisible here as well.

So the headline behavior from the PR description — "a position can run five hours past the Xetra cash close before an end-of-day exit" — has no test anywhere. `test_calendar.py` proves `paris_session_end_utc` returns 20:00/21:00 UTC, and `candle_source` passes it to the fetch, but no test walks the engine over a position that is still open at 17:30 and closes END_OF_DAY at 21:55.

Two ways out, either is fine:
- make `get_candles_in_window` honour `end` (generate bars up to it and slice), which makes the golden rows for G9HIC actually mean something and would probably shift its 569 lines; or
- leave golden as characterization-only and add one targeted test in `test_engine_impulsive.py` with candles past 17:30, asserting the exit lands on the last one.

The second is cheaper and covers the thing that matters.

---

api/services/backtest/calendar.py
```diff
@@ -64,22 +79,23 @@ def is_future_paris_date(
 
 
 def is_today_not_yet_closed(
-    d: datetime.date, now: Optional[datetime.datetime] = None
+    d: datetime.date,
+    market: Optional[Market] = None,
+    now: Optional[datetime.datetime] = None,
 ) -> bool:
```
RNiveau, line 91: Minor, but the same shape in all three functions: `market: Optional[Market] = None` falling back to `EUMarket()`.

`BacktestDefinition.market` is non-optional now, so every production call site already passes one — the default exists only for the tests and `market_fixture.py`. That makes it a default that can only ever be silently wrong: a future definition on a third market whose call site someone forgets keeps the 9:00–17:30 window and produces plausible-but-wrong results rather than a `TypeError`. Making `market` required and having the four test call sites pass `EUMarket()` explicitly costs nothing and removes the class.

Also here specifically: `market` was inserted as the second positional parameter, ahead of `now`. Every current caller uses `now=` so nothing is broken, but `is_today_not_yet_closed(d, some_datetime)` would now bind a datetime to `market` and fail somewhere downstream instead of at the call. A `*` before `now` pins it.

---

api/services/backtest/definitions.py
```diff
@@ -61,6 +66,28 @@
...
+    # trades the 9:00-22:00 CFD session, so a position can run five hours
+    # past the Xetra cash close before an end-of-day exit.
+    BacktestDefinition(
+        code="G9HIC",
+        name=Strategy.G9HIC.value,
+        display_name="GER40 Bougie de 9h (bougie impulsive)",
+        instrument="GER40.I",
+        market=EuCfdMarket(),
+        default_parameters=BacktestParameters(
+            stop_loss_points=150,
```
RNiveau, line 82: The comment is accurate — nothing reads `stop_loss_points` under an impulse stop, since `build_exit_chain` emits `Stop(only_when_armed=True)` and an armed stop is at entry. But it's still surfaced in `BacktestParametersResponse` and editable in the Backtest menu's override form, so a user can drag it from 150 to 30 on G9HIC and watch the results not move. That's already true of B9HWS, so this PR isn't introducing the problem — but it's now two of six definitions with a phantom control, which is about the point where the UI should probably grey it out (a `fixed_stop_applies` flag on the response would do it). Not for this PR; flagging so it doesn't get forgotten.

---

RNiveau (commented):
Reviewed the full diff, ran the suite and the linters locally against the branch.

**Verification:** `pytest tests/api/services/backtest tests/utils tests/services/test_candles_service.py tests/services/test_indicator_service.py` → 391 passed. `black --check`, `isort --check`, `flake8` clean; `mypy` clean apart from the four pre-existing missing-stub errors (`binance.error`, `aioboto3` ×3). The rest of `tests/api` doesn't collect in my container for an unrelated dependency reason (`annotated_types` missing under the py3.13 venv), so I couldn't run the router tests.

**What's good.** The rule really is one `Side` predicate plus one `ExitPolicy` composed in `build_exit_chain` — no engine code branches on which backtest is running, which is the property the #672 refactor was for and it held. The close-fraction test earning its place over a bare amplitude test is the right call and the docstring explains why. `closed_near_adverse_extreme` written multiplicatively to dodge the zero-range case is neat, and it's covered from four angles: direct unit tests, the 69/70 boundary pair, the exact-25% boundary, and the parametrised mirror-symmetry test. `__post_init__` rejecting half-configured, non-positive, out-of-range and structural-plus-impulse combinations at registration time is exactly where those checks belong. The two latent `end_minute > 59` `ValueError`s are real bugs found and fixed in passing, and the `model/market.py` extraction carries an honest explanation of the circular import that forced it. Golden diff is additive-only, so the "existing definitions byte-for-byte unchanged" claim checks out.

**The two I'd want answered before merge**, both inline:

1. `_fetch_daily_candles` now derives the daily regime series from `definition.market`, so G9HIC's `mm50_slope` / `adx14` / `overnight_gap` are computed over 13-hour CFD bars while every other definition uses 9-hour cash bars. That breaks the G9HIC-vs-G9HSL comparison the variant exists to make, and quietly changes what `overnight_gap` measures. It may well be deliberate; if so it deserves a line in the analysis doc, and if not it's a one-word fix.
2. The golden fixture ignores both the `end` bound and the `market` argument, so G9HIC's 569 new snapshot lines were generated over the 10:00–17:25 cash session. Nothing tests a position surviving past 17:30 to a 22:00 end-of-day exit — the behavior the PR description leads with.

Two minor notes inline as well (the `Optional[Market] = None` defaults in `calendar.py`, and the now-phantom `stop_loss_points` control in the UI).

---

api/services/backtest/service.py
RNiveau: Collateral, not deliberate — you're right, and it's reverted in b93111a. `_fetch_daily_candles` is back on `EUMarket()`, with the reasoning recorded in the docstring so the next person threading a market through doesn't "fix" it again.

Confirmed your reading of the mechanism: `build_daily_candles_from_h1` computes `num_h1 = close_hour - open_hour + (1 if open_minutes == 0)`, so EuCfdMarket gives 13 and EUMarket 9 — genuinely different daily bars, not just a different label. Your point 1 is the one that decides it: `DayResultSummary.mm50_slope` is documented as "lookahead-safe and config-independent", and a measure that changes with the definition being measured isn't config-independent. G9HIC vs G9HSL on the same instrument and dates is the entire purpose of the variant.

Point 3 is a good catch too, and pinning to EUMarket sidesteps it rather than fixing it — `build_daily_candles_from_h1` still doesn't range-check the splice. I've left that alone as out of scope here; happy to open an issue if you want it tracked.

---

tests/api/services/files/backtest_golden.json
RNiveau: Confirmed and fixed in b93111a — took your second option.

You're right on both mechanics: `get_candles_in_window` derives everything from `start.date()` and never looks at `end`, and `daily_candles` never looks at `market`. So the 569 golden lines were G9HIC-as-if-EUMarket, and the headline behaviour had no test. That also means the snapshot does **not** move with the daily-market revert from the other thread — which is itself the proof your point lands.

Four tests added in `test_engine_impulsive.py::TestCfdSession` / `TestSessionFetchWindow` (`m5_candle(n)` is 08:00 UTC + 5n, so offset 90 is exactly the 17:30 cash close):

- a position open past 17:30 closing `END_OF_DAY` at offset 143 — asserting `exit_time == 19:55 UTC`, i.e. 21:55 Paris;
- an impulse at offset 126 (20:30 Paris) stopping the position, so evening candles are demonstrably *evaluated*, not merely carried;
- the fetch window itself: `end == 20:00 UTC` for G9HIC against `15:30 UTC` for a cash definition, read off the `get_candles_in_window` call args.

That last one is the piece that was missing entirely — `test_calendar.py` proved the calculation and nothing proved it reached the fetch.

I left the golden fixture market-agnostic deliberately: making it honour `end` would shift G9HIC's rows and give the other five definitions a fixture whose behaviour depends on a bound they don't vary, for coverage the targeted tests now provide more legibly.

---

api/services/backtest/calendar.py
RNiveau: Both done in b93111a.

`market` is now required in all three functions, and the module docstring records why — a default that every production call site overrides can only ever be silently wrong. Four test call sites (`test_calendar.py` ×3 groups, `market_fixture.py` ×2) now pass `EUMarket()` explicitly, which reads better anyway: the cash-session expectation is stated rather than inherited.

`now` is keyword-only behind a `*`, so `is_today_not_yet_closed(d, some_datetime)` is a `TypeError` at the call site rather than a datetime bound to `market` and failing on `.close_hour` three frames later.

---

api/services/backtest/definitions.py
RNiveau: Agreed on all of it, and taking you at your word that it's not for this PR — no change here.

For the record so it isn't lost: the cleanest signal is a per-definition one rather than a name-based check, since "does a fixed stop distance apply" is already decided in one place — `build_exit_chain` emits `Stop(only_when_armed=True)` exactly when `structural_stop or impulsive_candle_points is not None`. A `fixed_stop_applies` field on `BacktestDefinitionResponse` derived from that same condition would keep the UI and the engine from drifting, rather than the frontend maintaining its own list of which codes have a phantom control.

Say the word and I'll open an issue with that sketch.

---

tests/api/services/backtest/test_engine_impulsive.py
```diff
@@ -0,0 +1,310 @@
...
+
+class TestCfdSession:
+    """FR-G12: the variant scans to 22:00 Paris, not the 17:30 cash close.
+
+    m5_candle(n) is 08:00 UTC + 5n minutes, i.e. 10:00 Paris + 5n in
+    summer, so offset 90 is exactly the 17:30 cash close and offset 143 is
+    the last 5-minute candle of the CFD session (21:55 Paris).
+    """
+
+    CASH_CLOSE_OFFSET = 90
```
RNiveau, line 252: `CASH_CLOSE_OFFSET` is never read — both tests use the literals 60/100/126 directly. Either drop it, or use it where it would actually document the intent (`m5_candle(self.CASH_CLOSE_OFFSET + 10, ...)` for the "after 17:30" candle reads better than a bare `100` and would stop being correct-by-coincidence if the base ever moved).

---

tests/api/services/backtest/test_definitions.py
```diff
@@ -68,6 +68,54 @@ def test_b9hws_is_the_wide_range_structural_variant(self):
...
+        # 21:00 + 60 minutes = a 22:00 Paris close.
+        assert (
+            definition.market.close_hour,
+            definition.market.end_minute,
+        ) == (
+            21,
+            60,
+        )
+
+    def test_only_the_impulsive_variant_left_the_cash_session(self):
```
RNiveau, line 120: Deriving the expectation from the registry instead of a second copy of the code list is the right shape, and this one does bite — I mutated `G9HIC` to `EUMarket()` and it failed alongside `test_g9hic_trades_the_cfd_session`.

Mild pushback on the sibling that went with it in a4424d4, though. `test_lists_the_hardcoded_backtests_in_order` wasn't purely a mirror: `BACKTEST_DEFINITIONS` order *is* the dropdown order in the Backtest menu, so the test pinned something user-visible that nothing else asserts now. The commit message's "could only fail by being out of date with the list it mirrors, never by catching a defect" is true of the membership half but not the ordering half — reordering the list today is a silent UI change. If you want to keep the two-file edit gone without losing that, asserting just the first entry (`list_definitions()[0].code == "B9H"`, i.e. "the default selection doesn't move") costs nothing and doesn't need touching when a definition is appended. Your call; it's test-only either way.

`assert list_definitions()` replacing `== 6` is fine — the test's stated purpose was construction-time validation, and the count was never part of that.

---

RNiveau (commented):
Re-reviewed `b5cad68..d7f527e`. Both blocking points are resolved, and resolved better than I asked for — this looks good to merge from my side.

**Verification:** 396 passed (was 391); `black`/`isort`/`flake8` clean; `mypy` clean apart from the same four pre-existing missing-stub errors. I also mutation-tested the three new guards rather than trusting that they pass:

- reverting `_fetch_daily_candles` to `definition.market` → `TestRegimeSeriesMarket::test_the_cfd_definition_also_measures_on_the_cash_session` fails, nothing else does;
- registering `G9HIC` on `EUMarket()` → `test_g9hic_trades_the_cfd_session` and `test_only_the_impulsive_variant_left_the_cash_session` fail;
- passing `EUMarket()` to `paris_session_end_utc` in `candle_source` → `TestSessionFetchWindow::test_the_impulsive_variant_scans_to_22h_paris` fails.

All three bite on exactly the mutation they're aimed at. Worth noting the second and third are caught by *different* tests, which is the helpers.py re-declaration convention working as designed — the registry test catches the shipped definition changing, the engine test catches the plumbing breaking, and neither masks the other.

**On the regime series:** pinning it to `EUMarket()` is the right resolution, and "a definition's market governs what it trades, never how it is measured" is a better articulation of the rule than my comment was. Encoding it as FR-G17b matters here because the golden fixture ignores the `market` argument entirely — the revert is invisible in the snapshot, so without `TestRegimeSeriesMarket` nothing at all would have stopped it coming back. The commit message says as much, which I appreciate.

**On the session coverage:** `TestCfdSession` plus `TestSessionFetchWindow` is the right split. The first proves the engine keeps evaluating past 17:30 (both that a position survives to a 21:55 exit and that an evening impulse still fires at 20:30 — the second is the one that would have caught a "carry but don't evaluate" bug); the second proves the definition's market actually reaches the Saxo fetch bounds, which is the part the golden fixture can't see. The 08:00 UTC + 5n arithmetic in the docstring checks out against `m5_candle`.

**Also fixed:** the `Optional[Market] = None` defaults are gone and `now` is keyword-only. `paris_reference_window_utc(d, EUMarket())` at the two fixture call sites is exactly the explicitness that was missing.

Two test-only nits inline, neither blocking — an unused constant, and a note about the ordering assertion dropped in a4424d4. Take or leave both.

---

tests/api/services/files/backtest_golden.json
RNiveau: Both nits taken in 255afbb.

**Ordering** — your pushback is right and my commit message overclaimed. I checked before encoding it: `Backtest.tsx:89` does `setSelectedDefinition(defs[0].code)` on load and renders the dropdown with `definitions.map`, so registry order really is user-visible in two ways, and after a4424d4 nothing asserted either. "Could only fail by being out of date" was true of the membership half and false of the ordering half; I collapsed the two when I deleted the test.

`test_the_default_menu_selection_does_not_move` now pins `list_definitions()[0].code == "B9H"` and nothing else, with the `Backtest.tsx` line referenced in the docstring so the reason survives. Appending a definition stays a one-file edit; moving the default fails. The membership half stays gone.

**`CASH_CLOSE_OFFSET`** — was declared and never read, which is worse than not having it since it reads as though the literals were derived from it. The boundary candles are now `CASH_CLOSE_OFFSET - 30`, `CASH_CLOSE_OFFSET + 10` and `CASH_CLOSE_OFFSET + 36`, so "36 candles past the cash close = 20:30 Paris" is stated rather than left for the reader to recompute.

Thanks for mutation-testing the guards rather than taking a green suite at face value — that the registry test and the engine test fail on *different* mutations is the property I was hoping for and hadn't checked.

682 passed locally, linters clean.

---

PR #680, opened 2026-07-28
feat: add two-lot impulsive backtest G9HICD with a TP1 trail (spec 025)

RNiveau:
A seventh definition, `G9HICD` — "GER40 Bougie de 9h (bougie impulsive, 2 lots)" — specified in [spec 025 §Addendum 3 and §Addendum 4](https://github.com/RNiveau/saxo-order/blob/claude/dax-backtest-impulsive-candle-61dj2p/specs/025-ger40-bougie-9h/spec.md) as FR-G24–FR-G35.

It is `G9HIC` at two lots:

- **TP1** = the level `G9HIC` exits at outright (H1 far level − 10).
- **TP2** = 100 points beyond TP1, i.e. **outside the H1 range** every other variant caps itself at (8190 against an 8000–8100 range).
- **TP1 filling arms break-even** on the runner, as `G9H` does.
- **After 50 points beyond TP1**, the runner's stop moves up again from break-even to **TP1** — a one-step trail locking the runner in to at least match the first lot.

Everything else is inherited from `G9HIC` unchanged — impulse stop, 70pt minimum H1 range, CFD session, 16:00 cut-off, two-loss cap — and `G9HIC` itself is untouched so it stays the single-lot control.

## Design

**No new exit code was needed for the two-lot part.** `build_exit_chain` already emits `DoubleTarget` for a two-lot definition and appends `ImpulsiveStop` for an impulse one, so `G9HICD` is the first definition to combine the two and neither had to be special-cased. Its chain is `Stop(only_when_armed)` → `DoubleTarget` → `ImpulsiveStop` → `ArmBreakEven` → `TrailToFirstTarget`.

**One refactor**: `LotModel.first_target_level(h1_high, h1_low)` becomes `targets(side, h1_high, h1_low, take_profit_level)`, returning both levels. The old signature could express neither half of the new model — it had no side (the H1 midpoint is direction-independent, the ordinary take-profit is not) and no say over the runner's level. `TwoLot` and the new `ExtendedTwoLot` share `TwoLotAccounting`; only where the targets sit differs. The engine now computes targets per side rather than once.

**The trail** is an arming policy at the tail of the chain beside `ArmBreakEven`, so a candle that spikes to the trigger and closes back below TP1 arms it without being stopped by it on that same candle. It measures from TP1, not entry: `entry + 50` is already the break-even trigger, and TP1 sits above the market at that point, so a stop there would fill instantly. A new `ExitReason.TRAILING_STOP` reports it — `break_even` would call a profitable exit flat, `take_profit` would claim a target was reached. The frontend renders reasons generically, so no UI change.

## Two things a reviewer should know

**1. The impulse stop can never close the runner.** TP1 arms break-even immediately and `ImpulsiveStop` stands down once armed, so it protects the pre-TP1 position only. That follows from the break-even-on-TP1 semantics, not from an implementation detail, and is recorded in the spec's Assumptions.

**2. `G9HICD` and `G9HIC` do not take the same set of trades.** I originally wrote SC-G15 claiming they did — that was wrong and is corrected in the spec rather than asserted away. Per breakout they accept and reject identically (TP1 is the ordinary take-profit `G9HIC` already checks against), but the runner holds the one-position-at-a-time slot until TP2, break-even or the close, so `G9HIC` can take entries `G9HICD` is still holding through. On the 42-day golden market this affects 4 days — 3 trades against 2, and once 6 against 4. **Compare points per day, not per trade.**

## Golden run

42 synthetic days. Only `G9HICD` is added; the other six definitions are byte-for-byte unchanged (verified key by key), including `G9HIC`.

| | `G9HIC` (1 lot) | `G9HICD` (2 lots) |
|---|---|---|
| trades | 63 | 58 |
| win / loss / BE | 24 / 13 / 26 | 22 / 14 / 22 |
| average win | 99.2 | 196.9 |
| average loss | 100.0 | 170.2 |
| final result | +1062.6 | +1948.8 |

The average win more than doubles while the average loss less than doubles, so the runner contributes beyond the pure size effect.

**The trail, measured on its own** (same definition, trail off vs on):

| | untrailed | trailed |
|---|---|---|
| win / loss / BE | 22 / 14 / 22 | 22 / 14 / 22 |
| average win | 207.4 | 196.9 |
| final result | **+2179.9** | **+1948.8** |

The trail **costs 231 points** with every count identical — it converts 9 exits to trailing stops and lowers only the average win, meaning those runners would have gone on to something better than TP1. That is the characteristic risk of a one-step trail, and a random-walk fixture with no trend persistence is the environment least favourable to trailing, so I would not read it as evidence against the rule. Flagging it rather than burying it: it is the first change in this sequence where the synthetic result moved clearly against the intent.

## Tests

- `test_engine_impulsive_double.py` (new, 17 cases): TP1→TP2, TP1→break-even, an impulse before TP1 closing both lots at twice the leg, an impulse after TP1 losing to the break-even stop, the +50 trigger before TP1, end of day with both lots and with the runner alone, the short mirror, entry parity with `G9HIC`, and the full trail set — fall back to TP1 as `trailing_stop`, no trail below the trigger, the trail never preventing TP2, arm-and-fall-back on one candle, no trail before TP1 fills, and the short trail mirror.
- `test_lots.py`: the `targets()` API and `ExtendedTwoLot`, including that it accounts points identically to `TwoLot`.
- `test_rules.py`: lot-model selection, the combined chain shape, and the trail policy's placement and trigger.
- `test_definitions.py`: the shipped fields, a test that `G9HICD` differs from `G9HIC` in *only* the two-lot fields, and the new `__post_init__` guards.

Two existing tests keyed on "exactly one impulsive definition" now derive their expectation from `impulsive_candle_points is not None` instead of a code, so they state the rule rather than a list.

754 passed, 10 skipped. `black`/`isort`/`flake8` clean; `mypy` at the four pre-existing missing-stub errors.

https://claude.ai/code/session_01DEcj6YEihziuZtkT6fPsP7

---
_Generated by [Claude Code](https://claude.ai/code/session_01DEcj6YEihziuZtkT6fPsP7)_


model/backtest.py
```diff
@@ -169,6 +207,24 @@ def __post_init__(self) -> None:
...
+                f"(definition {self.code!r})"
+            )
+        if (
+            self.trail_to_first_target_points is not None
+            and self.trail_to_first_target_points <= 0
+        ):
+            raise ValueError(
+                "trail_to_first_target_points must be positive - the trail "
+                "would otherwise arm the instant the first lot filled "
+                f"(definition {self.code!r})"
```
RNiveau, line 226: One guard missing from an otherwise complete set: `trail_to_first_target_points >= runner_extension_points` is a silent no-op, which is exactly what `__post_init__` exists to reject.

The trail arms at `TP1 + trigger`, and `TP2 = TP1 + extension`. `DoubleTarget` sits ahead of `TrailToFirstTarget` in the chain, so any candle that reaches `TP1 + extension` closes the position at take-profit before the trail policy is ever consulted. With `trigger == extension` the trail can never arm; with `trigger > extension` likewise. The definition would ship reporting results for a rule it never applied — the exact failure mode the docstring above cites.

Unlike the `TwoLot` case (where the TP1→TP2 distance is the H1 range and varies per day), both numbers are static here, so this is checkable at registration:

```python
if (
    self.trail_to_first_target_points is not None
    and self.runner_extension_points is not None
    and self.trail_to_first_target_points >= self.runner_extension_points
):
    raise ValueError(
        "trail_to_first_target_points must be short of the runner's "
        "target - the runner would otherwise take profit before the "
        f"trail could arm (definition {self.code!r})"
    )
```

---

tests/api/services/backtest/test_engine_impulsive_double.py
```diff
@@ -0,0 +1,322 @@
...
+            tp1_fill_candle(),
+            m5_candle(4, 8090, 8145, 8085, 8140),  # arms the trail
+            m5_candle(5, 8140, 8195, 8135, 8190),  # reaches TP2
+        ]
+
+        result = await run_impulsive_double(candles)
+
+        trade = result.trades[0]
+        assert trade.exit_reason == ExitReason.TAKE_PROFIT
+        assert trade.points == 250.0
```
RNiveau: The name claims more than the test shows: the trail *can* prevent TP2, on a candle that touches both.

Here TP1-retrace and TP2 fall on separate candles. When they collide on one candle, `Stop` leads the chain and the trailed stop wins. Ran it against this branch:

```python
candles = ger_entry_candles() + [
    tp1_fill_candle(),
    m5_candle(4, 8090, 8145, 8085, 8140),   # arms the trail
    m5_candle(5, 8140, 8195, 8080, 8190),   # dips to 8080 AND reaches TP2
]
# -> trailing_stop @ 8090.0, 150.0 points (untrailed: take_profit, 250.0)
```

That is FR-009 applied consistently (stop before target on the same candle), so I read it as correct, not a bug — but it is a second, untested way the trail costs points, and worth pinning down given the golden run already shows the rule at −231. Suggest renaming to something like `test_the_trail_does_not_prevent_tp2_on_a_later_candle` and adding the collision case asserting `trailing_stop`, so the convention is recorded rather than left to be rediscovered.

---

RNiveau (commented):
## Review

Adds a seventh backtest definition, `G9HICD` — `G9HIC` run at two lots with TP1 at the ordinary take-profit, TP2 100 points beyond it (outside the H1 range), break-even armed on the TP1 fill, and a one-step trail moving the runner's stop from break-even to TP1 after +50. Supporting changes: `LotModel.first_target_level` → `targets(side, …)` returning both levels, a new `ExtendedTwoLot` sharing `TwoLotAccounting` with `TwoLot`, a `TrailToFirstTarget` arming policy, `ExitReason.TRAILING_STOP`, and four new `__post_init__` guards.

**Verified locally** on `8e28d92`: 754 passed / 10 skipped, `black`/`isort`/`flake8` clean, `mypy` at the one pre-existing `aioboto3` stub error. The golden fixture diff is purely additive — one removed line, the `---` header — so the other six definitions are byte-for-byte unchanged as claimed.

### What holds up

- **Entry parity with `G9HIC` is real, not just asserted.** `is_valid_entry` rejects an entry at or past `first_target`, and for `ExtendedTwoLot` TP1 *is* the ordinary take-profit — so the guard `G9HIC` applies against its take-profit is the same comparison, against the same number. Passing TP2 as `take_profit_level` to `DirectionSearch` only loosens the weaker of the two checks. The per-day divergence and its cause (the runner holding the one-position slot) are documented in the PR body and in the test docstring rather than papered over, which is the right call.
- **The composition claim checks out.** `build_exit_chain` needed only the `TwoLot` → `TwoLotAccounting` isinstance widening for `DoubleTarget` + `ImpulsiveStop` to compose, and `test_a_two_lot_impulse_definition_composes_both` pins that shape.
- **Trail placement is right.** Tail of the chain beside `ArmBreakEven`, measuring from TP1 rather than entry, with `arm_level = TP1 + side.sign * trigger` — correctly mirrored for shorts, and `test_the_short_runner_trails_down_to_tp1` covers it. `stop_level` checking `trailed_to_first_target` ahead of `be_armed` is the correct precedence since the trailed stop is strictly better.
- **Deriving the two existing tests from `impulsive_candle_points is not None` instead of a code list** is a real improvement — they now state the rule rather than a snapshot of the registry.
- **The trail's −231 points is flagged rather than buried.** Agreed with the reading: a random-walk fixture with no trend persistence is the environment least favourable to a trail, and the counts being identical isolates the effect cleanly.

### Findings

Two inline comments — neither blocking:

1. **`model/backtest.py`** — missing guard: `trail_to_first_target_points >= runner_extension_points` ships the trail as a silent no-op (`DoubleTarget` precedes `TrailToFirstTarget`, so TP2 always closes first). Both values are static, so it is checkable at registration, and every neighbouring guard exists for exactly this failure mode.
2. **`test_the_trail_never_prevents_tp2`** — the name overstates the property. On a single candle that both retraces to TP1 and reaches TP2, the trailed stop wins (confirmed: `trailing_stop @ 8090, 150.0` vs `take_profit, 250.0` untrailed). Consistent with FR-009, but it is a second untested way the trail costs points; worth renaming and pinning.

### Smaller notes

- `frontend` needs no change — `formatReason` is a plain `replace(/_/g, ' ')`, so `trailing_stop` renders as "trailing stop". Nothing else in `api/` or `frontend/src` switches on `ExitReason`. Confirmed.
- `TrailToFirstTarget.resolve`'s `first_target is None` check is unreachable given the `first_target_taken` guard above it, but it is what keeps `mypy` happy on the `+` — fine as is.
- `build_exit_chain` constructs a lot model purely for an `isinstance` check, then `BacktestService` builds another. Pre-existing, and cheap, but the widening to `TwoLotAccounting` makes the throwaway construction slightly more visible.

Nothing here blocks merge — item 1 is the one I would take before this ships, since it is the same class of silent no-op the rest of `__post_init__` already defends against. Note the branch is behind `main`.

---

model/backtest.py
RNiveau: Taken in 9e54e11 — you're right, and it's the same class of silent no-op the rest of `__post_init__` exists to catch. Added essentially your snippet, with the chain ordering named in the message so the reason survives:

```
"trail_to_first_target_points must be short of the runner's target -
 DoubleTarget precedes the trail in the chain, so the runner would take
 profit before the trail could ever arm (definition {code})"
```

Your distinction from the `TwoLot` case is the part that makes it checkable and worth stating: there the TP1→TP2 distance is the H1 range and varies daily, so no registration-time check is possible; here both numbers are fixed on the definition, so a trigger that can never fire is knowable at import. Covered by a parametrised test at `trigger == extension` and `trigger > extension`, and FR-G35 now carries the "strictly less than" requirement.

---

tests/api/services/backtest/test_engine_impulsive_double.py
RNiveau: Reproduced your snippet verbatim before changing anything — `trailing_stop @ 8090.0, 150.0`, exactly as you had it. Fixed in 9e54e11.

The test is now `test_the_trail_does_not_prevent_tp2_on_a_later_candle` (separate candles, unchanged assertions), and `test_the_trailed_stop_wins_when_one_candle_hits_both` pins the collision with your candles, asserting `trailing_stop` at 8090 for 150 points.

Agreed it is correct rather than a bug — it is FR-009's stop-before-target convention applied to the trailed level, the same rule that makes an ordinary stop beat a target on one candle. But your framing is the useful bit: it is a *second* way the trail costs points, distinct from cutting short a runner that would have gone further, and the golden run already has the rule at −231. Both costs are now in the spec's edge cases so the −231 has a mechanism attached rather than being an unexplained number.

Also merged `main` in — the branch was behind by #679.

---

PR #681, opened 2026-07-28
refactor: key the backtest candle cache on instrument and session, with a migration

RNiveau:
## Why

The raw-candle cache was keyed on the definition code (`{code}:{instrument}:v1`), so every strategy on an instrument stored its own copy of identical Saxo candles — three copies of the FRA40.I series, three of GER40.I. The cached bytes only depend on the instrument and the session window the candles are fetched over.

## What changed

**The key** (`api/services/backtest/candle_source.py`) is now `{instrument}:{session window}:v2`, e.g. `FRA40.I:0900-1730@Europe/Paris:v2`. The session window stays in the key because `G9HIC` runs `EuCfdMarket` and fetches 5-minute candles to 22:00 rather than 17:30; the timezone is part of it because the hours are exchange-local and `market_in_utc` resolves them against `market.timezone`, so identical local hours elsewhere would cover a different UTC window. Six definitions collapse to two live keys, plus one for the GER40 CFD session.

**Partial entries.** The `is_below_min_range` short-circuit (FR-033) was writing `has_data=True` with an empty `m5_candles` for `B9HWS`/`G9HIC`. Under a shared key that would have told `B9H` the day had no 5-minute candles. Entries now carry `m5_fetched`; a partial one is served as-is to a below-min-range definition and completed for anyone else — a 5-minute-only fetch reusing the cached H1. The partial write is conditional on the key being absent, so a run racing another on the same day cannot downgrade a complete entry.

**The migration** — `k-order internal migrate-backtest-cache [--dry-run]` (`api/services/backtest/cache_migration.py`). Planning is a pure function over the scanned items, so `--dry-run` prints real counts. It writes every new entry before deleting any old one and only deletes a row whose replacement actually landed, so an interrupted run leaves duplicated data, never missing data, and re-running finishes the job. Collisions (N old rows → 1 new key) are resolved on `(has_data, m5_fetched, len(m5_candles))`. `m5_fetched` is reconstructed for v1 items from the writing definition's threshold versus the cached H1 range. Rows whose definition code no longer exists are reported as orphans and left in place; rows that don't match the v1 key shape are skipped — neither is ever mis-mapped.

The DynamoDB hash key attribute keeps the name `definition_code` (it now holds a cache key, documented at `client/aws_client.py:974`): renaming a key attribute forces a table replacement, which would destroy exactly the data being migrated. Pulumi is unchanged.

## Specs

`specs/021-backtest-menu-hardcoded/` recorded the per-definition key as "a deliberate simplicity tradeoff, not a defect" — that decision is marked superseded in `spec.md`, with `data-model.md`, `plan.md` and `research.md` §8 updated. The planning docs keep their original rationale with a dated superseding note rather than being rewritten. `specs/025-ger40-bougie-9h/plan.md`'s cache note is corrected: the CFD variant is still isolated, now by its session window rather than its code.

## Testing

734 tests pass (14 new, covering the shared key, session/timezone separation, partial-entry completion, and the migration's collision, orphan and failure paths). black / isort / flake8 clean; mypy unchanged at its 4 pre-existing missing-stub errors.

**Not done:** the migration has not been run against the real table — it needs AWS credentials and it deletes rows. Worth a `--dry-run` first. Note also that keeping `v2` rather than bumping to `v3` for the timezone fix is only safe because the migration hasn't run yet; once it has, a key-shape change needs a new version and a second pass.

https://claude.ai/code/session_01Evsvj1TTFZhzUqGzFRz9U5

---
_Generated by [Claude Code](https://claude.ai/code/session_01Evsvj1TTFZhzUqGzFRz9U5)_


RNiveau:
## Review

Reviewed against the branch locally. The PR's own claims all check out: **734 tests pass** (10 skipped), black / isort / flake8 clean, mypy at the 4 pre-existing missing-stub errors.

### The core design is right

I checked the premise carefully, since a wrong shared key silently corrupts backtest results:

- `session_key` is built from exactly `open_hour` / `open_minutes` / `close_hour` / `end_minute` / `timezone` — and those are precisely the fields `paris_reference_window_utc` and `paris_session_end_utc` consume (`api/services/backtest/calendar.py:44-76`). Nothing else in the fetch is definition-dependent: both `_fetch_*` helpers take only `(instrument, start, end)`. The abstraction holds.
- Normalizing `close_hour * 60 + end_minute` is subtle and correct: `EuCfdMarket`'s `close_hour=21, end_minute=60` and a hypothetical `22/0` both render `2200` and *do* produce the same UTC bound, so that collision is a feature.
- Including `timezone` is the conservative choice — two tz names with identical offsets get separate entries, wasteful but never wrong.
- `_was_m5_fetched`'s reconstruction (`h1_range > min_h1_range_points`) is the exact negation of `is_below_min_range`'s `<=`. Correct.
- Write-before-delete, `_is_covered` gating each delete on its replacement landing, `only_if_absent` on the partial write, orphans left in place rather than guessed at — the safety posture is well thought through.

### Migration write ordering — scoped to a single run

`CacheMigration.apply` writes with `only_if_absent=False`, and `build_plan` ignores existing v2 rows entirely (they land in `skipped`). So a v1 row can overwrite a v2 entry that is already there. Two variants, with different weight given the stated plan to run this exactly once:

**Re-run variant — effectively moot, with one wrinkle.** It needs a leftover v1 row, which only exists if a run didn't finish cleanly. But that is exactly the case where you *would* re-run: `apply` is built to leave v1 rows behind on a failed write, and its docstring says "the fix is to run it again". So "only once" holds precisely when the migration succeeds, and stops holding in the one situation that would make you reach for a second run. Practically: check the `failed` count in the output. If it is 0, this is moot.

**Concurrency variant — survives, but small.** The scan happens up front; a backtest writing a v2 entry mid-run gets overwritten. Realistically that means a completed entry replaced by a partial one, which `_complete_m5` heals on the next read at the cost of one 5-minute refetch. The `has_data=False` case needs every v1 row for that day to be a no-data marker while the app simultaneously found real data — rare, and worth knowing about rather than engineering around.

Free mitigation: run it when no backtest is executing. That closes the window at no cost. No code change needed for a clean single run.

### Test coverage gap

The client-layer primitives added here have no direct tests — `tests/client/test_aws_client_backtest_cache.py` still has its original two. Specifically untested: `only_if_absent`'s `ConditionalCheckFailedException` handler, `scan_backtest_candles`'s pagination loop, and `delete_backtest_candles`.

The conditional-write handler is the one I would actually add coverage for. Unlike the migration, it runs on **every below-min-range day, forever** — it is the sole guard against a partial entry clobbering a complete one, and that gap is independent of how many times you migrate. The service-layer coverage is otherwise strong; the session/timezone-collision cases in `test_calendar.py` are the right ones to have.

### Minor

- `saxo_order/commands/internal.py:369` — "v1 entries found" prints `len(plan.deletes)`, which excludes orphans; orphans *are* v1 entries found. "already migrated" prints `skipped`, which also counts malformed rows and rows with a non-string key.
- `MigrationResult.failed` conflates failed writes with failed deletes — different consequences (data not migrated vs. a duplicate left behind), same counter.
- `api/services/backtest/candle_source.py:88` — bare `assert` on a production path, stripped under `-O`. The invariant does hold via `_cache_hit`, so this is style only.
- `_complete_m5`'s `SaxoException` path returns `has_data=True, m5_candles=[]` with `m5_fetched` defaulting to `True`, so a transient failure looks like a complete no-signal day to the strategy. It is not cached and it mirrors `_fetch_and_store`'s existing behaviour, so this is consistent rather than a regression — just noting the silent NO_TRADE.

### Ops notes

The PR is honest that the migration has not been run. Two things worth having in the runbook:

1. Run with backtests quiesced (see above).
2. Reverting this PR after the migration lands drops `CACHE_SCHEMA_VERSION` back to 1 with the v1 rows already deleted — not data loss, but a full cold refetch from Saxo.

Suggested sequence: `--dry-run` first, confirm the counts match expectations, run for real with backtests quiesced, then check that `failed` is 0.

### Verdict

**Approve.** Keeping the `definition_code` attribute name to avoid a table replacement is the right call and well documented, and the shared-key premise holds up under scrutiny. The client-layer test for the conditional write is the only thing I would genuinely like to see added, and it is not a merge blocker.

---

RNiveau:
Thanks — addressed in 57597f8. 747 tests pass, lint clean, mypy unchanged.

**Migration write ordering.** Fixed rather than left to the runbook. `build_plan` no longer ignores current-schema rows: they contend for their target slot alongside the v1 rows, and a v1 row only overwrites one if it ranks strictly better (ties go to the existing row, so no pointless rewrite). A slot an existing row already covers needs no write, but the v1 rows feeding it are still deleted — tracked as `already_present` and seeded into `apply`'s covered set so the delete gating still holds.

That closes the re-run variant you identified, which is the one that mattered: the recovery path for a partly-failed run is to run again, so "runs exactly once" stops holding in exactly the case where a stale v1 copy could win. The concurrency variant survives — the scan is still up front — so the quiesce advice stays, now in the runbook alongside the revert note.

**Client-layer tests.** Added, agreeing with your reasoning that the conditional-write handler is the one that earns it — it runs on every below-min-range day rather than once. Covered: the condition expression being set, the default write staying unconditional, `ConditionalCheckFailedException` returning cleanly, and a non-condition `ClientError` still propagating as `DynamoDBOperationError`. Plus the scan pagination loop and delete.

**Minor.** All four taken: `v1_entries` now includes orphans and the "already migrated" line no longer absorbs malformed rows; `MigrationResult` splits `write_failures` from `delete_failures`, with the CLI printing the right follow-up for each; the bare `assert` is now a real check.

**Not changed:** `_complete_m5`'s `SaxoException` path. Agreed it is consistent with `_fetch_and_store` rather than a regression, and making the silent NO_TRADE visible is a change to how transient Saxo failures surface generally — worth its own change, not this one.

On the ops sequence — `--dry-run`, confirm counts, run quiesced, check failures is 0 — that is now in `data-model.md`, and the failure count is split so a non-zero write count and a non-zero delete count give different instructions.

---

PR #682, opened 2026-07-29
fix: correct combo indicator band comparisons and speed up macd0lag

RNiveau:
## Summary

Fixes four defects found while analysing the `combo` indicator, plus a magic-number cleanup.

### 1. `bb25_1` was not shifted
`bb25_1 = bollinger_bands(candles, 2.5)` was identical to `bb_last` — no slicing. The "candle -1" branch of `price_within_bb` therefore compared the *previous* candle against the *current* candle's 2.5σ band while pairing it with the correctly-shifted `bb20_1`. Now `bollinger_bands(candles[1:], 2.5)`.

### 2. Inconsistent `price_within_bb` tolerances
The two branches used four different values and neither mirrored the buy side:

| | outer band (2.5σ) | inner band (2.0σ) |
|---|---|---|
| BUY, both branches | 0.05% | 0.5% |
| SELL, candle-1 branch | 0.5% | 0.05% |
| SELL, candle-0 branch | 0.5% | 0.5% |

Unified at 0.5% (`COMBO_BB_TOLERANCE`). The sell branches are rewritten in the same `low < close < high` orientation as the buy ones — the reversed orientation is what made the asymmetry hard to see.

### 3. Buy-side MA50 proximity clause was a no-op
```python
candles[0].higher > ma50_last + margin_variation_ma50
```
The preceding clause already requires `close > ma50 + margin`, and `higher >= close`, so both `higher` clauses were unconditionally true — the MA50 conjunction collapsed to just the close checks. The BB clause directly above uses `lower` for the same "even the extreme is clear of it" logic, and the sell mirror correctly uses `higher`. Now uses `lower`.

The buy/sell copies of this block were 8-line near-duplicates differing only in comparison direction and which extreme they read — exactly where the bug was hiding. Extracted into `is_far_from_supports()`, which takes a `Direction`.

### 4. `macd0lag` performance
`_macd0lag` ran 81 times, each rebuilding ~114 EMAs over heavily overlapping suffixes of the same list — ~9,300 EMA passes per call, per asset, per alerting run. An `(offset, period)` cache brings it to ~275. Same function, same inputs, so **output is bit-identical**: the golden expectations (`-36.07337`, `113.46834`) pass untouched. `tests/services/test_indicator_service.py` runs 2.39s → 0.38s.

### 5. Magic numbers extracted
`COMBO_MA50_SLOPE_MIN/STRONG`, `COMBO_BB_FLAT_SLOPE_MAX`, `COMBO_BB_BREACH_TOLERANCE`, `COMBO_BB_TOLERANCE`, `COMBO_ATR_BB_MARGIN`, `COMBO_ATR_MA50_MARGIN`, `COMBO_STRONG_SIGNAL_MIN`, alongside the existing `MM50_TOUCH_*`. The flat-bands test is computed once as `both_bb_flat` instead of repeated per branch.

## Test coverage caveat

**No existing golden expectation changes.** I verified why rather than treating that as proof of correctness — a differential run over all 13 combo fixtures shows:

- `price_within_bb`: unchanged on every fixture. The 0.9995 → 0.995 widening only matters for a close within 0.45% of a band, and none land there.
- The MA50 fix flips its clause on `combo_buy_h4_dax.obj` (`True` → `False`), but that fixture's outer BB clause is `False`, so the MA50 leg is short-circuited and the outcome is identical.

So the recorded fixtures don't exercise the fixed paths at the output level. The 7 new `TestIsFarFromSupports` tests close that gap for fix 3. **Fixes 1 and 2 remain without a fixture that distinguishes old from new** — that needs a case recorded from real data with price sitting within ~0.5% of a band, which I did not want to fabricate.

## Test plan

- `poetry run pytest` — 798 passed, 10 skipped
- black / isort / flake8 / mypy clean (pre-commit hooks passed)

### DIFF
--- a/services/indicator_service.py
+++ b/services/indicator_service.py
@@ -103,6 +103,15 @@ def mobile_average(candles: List[Candle], period: int) -> float:
 MM50_TOUCH_PROXIMITY = 0.01
 MM50_TOUCH_SLOPE_MIN = 3.0
 
+COMBO_MA50_SLOPE_MIN = 3.0
+COMBO_MA50_SLOPE_STRONG = 10.0
+COMBO_BB_FLAT_SLOPE_MAX = 5.0
+COMBO_BB_BREACH_TOLERANCE = 0.001
+COMBO_BB_TOLERANCE = 0.005
+COMBO_ATR_BB_MARGIN = 0.3
+COMBO_ATR_MA50_MARGIN = 0.1
+COMBO_STRONG_SIGNAL_MIN = 4
+
 
 def mm50_touch(candles: List[Candle]) -> Optional[Dict[str, float]]:
     if len(candles) < 60:
@@ -138,6 +147,32 @@ def containing_candle(candles: List[Candle]) -> Optional[Candle]:
     return None
 
 
+def is_far_from_supports(
+    candles: List[Candle],
+    band: float,
+    ma50: float,
+    margin_band: float,
+    margin_ma50: float,
+    direction: Direction,
+) -> bool:
+    """
+    True when the last two candles sit clear of both the 2.0 bollinger band
+    and the ma50: the pullback never came close enough to be tradable.
+    The candle extreme facing the level is used - the low for a buy, the
+    high for a sell - so a wick reaching the level counts as a touch.
+    """
+    closes = [candle.close for candle in candles[:2]]
+    if direction == Direction.BUY:
+        values = closes + [candle.lower for candle in candles[:2]]
+        return all(value > band + margin_band for value in values) and all(
+            value > ma50 + margin_ma50 for value in values
+        )
+    values = closes + [candle.higher for candle in candles[:2]]
+    return all(value < band - margin_band for value in values) and all(
+        value < ma50 - margin_ma50 for value in values
+    )
+
+
 def combo(candles: List[Candle]) -> Optional[ComboSignal]:
     logger = Logger.get_logger("combo")
     logger.debug(
@@ -146,7 +181,7 @@ def combo(candles: List[Candle]) -> Optional[ComboSignal]:
     ma50_last = mobile_average(candles, 50)
     ma50_first = mobile_average(candles[10:], 50)
     bb_last = bollinger_bands(candles, 2.5)
-    bb25_1 = bollinger_bands(candles, 2.5)
+    bb25_1 = bollinger_bands(candles[1:], 2.5)
     bb_first = bollinger_bands(candles[2:], 2.5)
     bb20 = bollinger_bands(candles, 2.0)
     bb20_1 = bollinger_bands(candles[1:], 2.0)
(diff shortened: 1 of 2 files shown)


RNiveau (commented):
## Review

I verified each of the five claims against the diff. The refactors are behaviour-preserving where they claim to be, the three real fixes are correct, and the write-up is honest about the coverage gap. A few things worth addressing before merge, none of them blocking on their own.

### Verified as correct

**Fix 1 (`bb25_1`)** — real bug, correct fix. `bb25_1` was `bollinger_bands(candles, 2.5)` while its partner `bb20_1` was already `candles[1:]`, so the candle‑1 branch mixed a shifted inner band with an unshifted outer one. Ordering still holds after the fix (`bb25_1.bottom < bb20_1.bottom`, `bb20_1.up < bb25_1.up`), so both intervals remain well-formed.

**Fix 3 (MA50 leg)** — the diagnosis holds. `close > ma50 + margin` combined with `higher >= close` makes `higher > ma50 + margin` unconditionally true, so the whole MA50 conjunction reduced to the two close checks. Switching to `lower` (buy) matches both the BB clause above it and the sell mirror. Note this is a genuine behavioural change: the discard condition gets *stricter*, so `combo` will emit signals it previously suppressed. That's the intent, but it's the one change in this PR that can alter live alerting volume — worth a look at how often it fires before this ships.

**Pure refactors — checked equivalent, no behaviour change:**
- `(x < -5 or x > 5)` → `abs(x) > 5`, and `5 > x > -5` → `abs(x) < 5`: equivalent (both false at exactly ±5, as before).
- `signal > 3` → `signal >= COMBO_STRONG_SIGNAL_MIN` (4): identical over integers. Max attainable signal is 5.
- `_macd0lag` list lengths: old built `[ma] + [ema(candles[i:]) for i in 1..3p-1]` = `3p` entries; new builds `[_ema(offset+i) for i in range(3p)]` = `3p` entries, with `i=0` reproducing the old head element. The guard `len(candles[i:]) < 4L` → `len(candles) - offset < 4L` is the same predicate. `candles` is fixed for the lifetime of the cache and `exponentiel_mobile_average` is pure, so the "bit-identical" claim is sound — the golden values passing unchanged is real evidence here, not a coincidence.

### Worth changing

**1. The BB leg of `is_far_from_supports` is effectively untested.** `test_buy_close_below_the_band_is_not_far` uses `close=962` against `MA50=1000`, so the MA50 leg fails too — the test passes even if the band conjunction is deleted entirely. And with the class's stated geometry (band 960 below ma50 1000) the band leg can *never* be the deciding one for a buy, which the docstring acknowledges. Given that fix 3 was itself a leg that had silently collapsed to a no-op, leaving the sibling leg with no test that can distinguish it from `True` is the wrong place to economise. Suggest one buy case with the band above the MA50 (unrealistic geometry, but it isolates the leg) and one sell case where `candles[1].higher` — not `candles[0].higher` — is what pierces.

**2. `is_far_from_supports` silently accepts fewer than two candles.** `candles[:2]` degrades to a one- or zero-element check where the old inline code would have raised `IndexError` on `candles[1]`. On an empty list both `all()` calls return `True`, i.e. "far from supports" → signal discarded. `combo` can't reach that state today (macd0lag needs ~235 candles), but the function is now public and independently testable, so a `len(candles) < 2` guard is cheap insurance.

**3. Naming.** For the sell direction the levels are resistances, not supports — `is_far_from_levels` or similar would read better. Minor.

**4. `Dict[tuple, float]`** — bare `tuple` loses the key shape. `Dict[Tuple[int, int], float]` documents the `(offset, period)` contract the comment above it describes.

### Adjacent finding, not introduced here

While auditing the band comparisons: `bb_first = bollinger_bands(candles[2:], 2.5)` is **2** candles back from `bb_last`, but the slope is taken over an x-distance of **3**:

```python
bbh_slope = slope_percentage(0, bb_first.up, 3, bb_last.up)
```

Compare `ma50_slope`, where `candles[10:]` correctly pairs with a distance of `10`. If this is an off-by-one, both `bbh_slope` and `bbb_slope` are being scaled by ~2/3, which feeds straight into `COMBO_BB_FLAT_SLOPE_MAX` and the `both_bb_flat` signal — the bands read flatter than they are. Pre-existing and out of scope for this PR, but this PR is exactly the audit that should catch it. Either `bb_first = bollinger_bands(candles[3:], 2.5)` or a distance of `2`, depending on which was intended.

### On the coverage caveat

The differential analysis over the 13 fixtures is the right thing to have done, and stating plainly that fixes 1 and 2 have no fixture that distinguishes old from new is the right call — better than a fabricated fixture. Worth tracking as a follow-up so it doesn't get lost: the fixed paths need a recording from real data with price sitting within ~0.5% of a band.

---

PR #689, opened 2026-07-30
refactor: collapse the mirrored combo branches and defer macd0lag

RNiveau:
The buy and sell halves of combo() were exact mirrors, ~137 of its 168
lines. Two of the three defects fixed in #682 were one-sided drift
between halves meant to be identical (.higher used against the ma50
support on the buy side, 0.995 vs 0.9995 on the sell side), so the
duplication is what let them hide. Both directions now run through one
_combo_for_direction body where sign flips every comparison.

macd0lag was computed up front but feeds a single scoring criterion, so
it moves behind a lazy property on the new _ComboContext. It needs 235
candles where the rest of combo needs 60, and the eager call meant a
short history raised SaxoException instead of returning None - an
exception that run_detection_for_asset catches after six detectors have
run and before store_alerts, silently discarding every alert already
collected for that asset. Rejected candle sets are also 20-25x faster
now, unchanged when a signal is actually scored.

Verified behaviour-preserving by diffing the old and new implementations
over 508 sliding candle windows from the recorded fixtures: no
divergence, and the only two differences are short histories that used
to raise and now correctly return None.

Also renames the bollinger locals to say which band and which offset they
hold (bb_last/bb25_1/bb_first -> bb25/bb25_previous), makes
is_price_within_bands take outer/inner as keyword-only since silently
swapping two adjacent floats is the shape of the bug just fixed, types
macd0lag's return, and corrects its documented candle minimum to 235.

Claude-Session: https://claude.ai/code/session_01F5cqHxv1EJW3Br3u8GJ6P3

### DIFF
--- a/saxo_order/commands/alerting.py
+++ b/saxo_order/commands/alerting.py
@@ -43,12 +43,23 @@ def _safe_detect(
     (IndexError on an asset that returned no candle) escaped the handler and
     aborted the entire scan. A detector that cannot run is not an error for
     the others, and never a reason to drop what has already been found.
+
+    Insufficient history is routine and logs at warning; anything else is a
+    bug that must stay visible, so it logs at error with a traceback rather
+    than being flattened into the same routine-looking line.
     """
     try:
         return detector()
-    except Exception as e:
+    except (SaxoException, IndexError) as e:
         logger.warning(
             f"{alert_type.value} skipped for {asset_description}: {e}"
+        )
+        return None
+    except Exception as e:
+        logger.error(
+            f"{alert_type.value} failed unexpectedly for"
+            f" {asset_description}: {e}",
+            exc_info=True,
         )
         return None
 
--- a/saxo_order/commands/alerting.py
+++ b/saxo_order/commands/alerting.py
@@ -175,7 +175,7 @@ async def run_detection_for_asset(
     country_code: Optional[str],
     exchange: str,
     asset_description: str,
-    saxo_uic: Optional[str],
+    saxo_uic: Optional[str | int],
     saxo_client: SaxoClient,
     dynamodb_client: DynamoDBClient,
 ) -> List[Alert]:
--- a/saxo_order/commands/alerting.py
+++ b/saxo_order/commands/alerting.py
@@ -3,7 +3,8 @@ import datetime
 import json
 import os
 import time
-from typing import Dict, List, Optional, Tuple
+from functools import partial
+from typing import Callable, Dict, List, Optional, Tuple, TypeVar
 
 import click
 from click.core import Context
@@ -24,6 +25,32 @@ from utils.helper import build_daily_candles_from_h1
 from utils.logger import Logger
 
 logger = Logger.get_logger("alerting")
+
+T = TypeVar("T")
+
+
+def _safe_detect(
+    alert_type: AlertType, asset_description: str, detector: Callable[[], T]
+) -> Optional[T]:
+    """
+    Run one detector in isolation, returning None when it cannot run.
+
+    Each detector has its own history requirement - combo needs 235 candles,
+    double_inside_bar 3, double_top a non-empty list - and they used to share
+    one try/except wrapping the whole chain plus the DynamoDB write. So the
+    most demanding detector decided how much history an asset needed before
+    *any* of its alerts were persisted, and an error type nobody had listed
+    (IndexError on an asset that returned no candle) escaped the handler and
+    aborted the entire scan. A detector that cannot run is not an error for
+    the others, and never a reason to drop what has already been found.
+    """
+    try:
+        return detector()
+    except Exception as e:
+        logger.warning(
+            f"{alert_type.value} skipped for {asset_description}: {e}"
+        )
+        return None
 
 
 def _parse_asset_code(code: str) -> Tuple[str, Optional[str]]:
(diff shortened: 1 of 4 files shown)


RNiveau:
## Review

I re-verified the two central claims independently rather than taking the description at its word, and both hold. One thing will fail CI though.

### What I checked

**Behaviour preservation.** Ran the pre-refactor `combo` (from `main`) against the new one over every `tests/services/files/*.obj` fixture, sliding the window one candle at a time — 3484 windows, comparing `price`/`direction`/`strength`/`has_been_triggered` and the full `details` dict:

```
identical: 684   (67 of which produced an actual ComboSignal)
old_raised_short_history: 2800   (old raised SaxoException, new returns None, all len < 235)
real divergences: 0
```

I also walked the sign algebra of each mirrored comparison by hand (ma50 side, the `(1 - sign * TOLERANCE)` band breach, `ma50_over_bb`, `strong_ma50`, the macd comparison, trigger/pending levels). All six flip correctly, and the `details` key insertion order is unchanged, which matters since it is serialised into the alert.

**Performance.**

```
scored signal (combo_buy_daily_cac, 250 candles): old 8.5ms -> new 8.7ms  (1.0x)
rejected flat set (300 candles):                  old 10.4ms -> new 0.4ms (26x)
```

Collapsing 137 duplicated lines into one sign-flipped body is clearly the right call here — the two defects in #682 were one-sided drift, and this removes the place drift can hide.

### Blocking: mypy fails

`tests/saxo_order/commands/test_alerting.py:465` — `saxo_uic=12345` is an `int`, the parameter is `str | None`. The existing tests get away with this because they are unannotated (mypy skips their bodies); the new `_run` helper is annotated `-> tuple`, so its body is checked. Running the exact CI command:

```
$ poetry run mypy --explicit-package-bases --ignore-missing-imports --exclude pulumi .
tests/saxo_order/commands/test_alerting.py:465: error: Argument "saxo_uic" ... has incompatible type "int"; expected "str | None"
Found 1 error in 1 file
```

`saxo_uic="12345"` fixes it. The second call site (~486, in `test_a_failing_store_does_not_raise`) has the same mismatch and is only unflagged because that function is unannotated — worth fixing both.

`black`, `isort`, `flake8` are clean and all 115 tests in the two touched files pass.

### Worth discussing

**`_safe_detect` catches bare `Exception` at `warning`.** The isolation is right, but the level flattens two very different things. A missing-history `SaxoException` is expected and routine; a `TypeError` from a future refactor of any detector is a bug that would now log one warning per asset and let the scan report success. Consider warning for the expected shortage cases (`SaxoException`, `IndexError`) and `error` for anything else, so a genuine regression still stands out in the Lambda logs.

**`combo` can still raise for 60 ≤ len(candles) < 235.** Deferring `macd0lag` only helps candle sets rejected by the ma50/BB gates — once a direction is actually scored, `criteria["macd"]` reaches it and an asset with ~200 candles and a real signal still raises, now silently losing just its COMBO alert. Much better than losing the whole asset's alerts, so not a blocker, but the underlying gap is still there. Making the `macd` criterion `False` on short history (rather than raising) would close it. Relatedly: the docstring minimum is now correctly 235, but the runtime guard still logs `needed 104`, which is what sent this down the wrong path in the first place.

### Nits

- `TestComboDefersMacd0lag`'s docstring says "needs 233 candles"; everywhere else says 235.
- `test_short_history_never_raises` asserts only `isinstance(alerts, list)`. With flat closes the ma50 slope is 0, so even the `234` case bails at the ma50 gate — none of the parameters exercise "direction scored with short history", which is the path that still raises. A case built from a truncated signal fixture would pin that behaviour.
- `_ComboContext.__init__` still eagerly computes `bb20`, `bb20_previous` and the ATR margins, none of which are read on the `neither_bb_flat` rejection path. The 26x win already covers the expensive part, so this is cosmetic — but they are the same shape of candidate as `macd0lag` if you want the rejection path leaner still.

Nothing here changes my read that the refactor is sound. Fix the mypy error and it is good to merge; the `_safe_detect` log level is the one I would push on.

---

RNiveau:
Thanks — the independent differential (3484 windows vs my 508) and the hand-check of the sign algebra are exactly the verification I'd want on this. All addressed in `d6f1023`, CI green. Two of them I resolved differently than suggested, with reasoning.

### mypy — fixed, but the other way round

Fixed in `212fdcc` before this review landed, by widening the annotation to `Optional[str | int]` rather than changing the tests to `saxo_uic="12345"`.

The annotation was the wrong one. Both production call sites pass the Saxo API's integer uic — `asset.get("saxo_uic")` at `alerting.py` and `asset_info.get("Identifier")` in `api/services/alerting_service.py` — and `get_historical_data` downstream already declares `str | int`. mypy never flagged those because a dict `.get()` is `Any`. Making the tests pass a string would have made them green while encoding a type production never sends, and left the second call site you spotted needing the same false change. Widening the annotation fixes both at once and makes the signature honest.

### `_safe_detect` log level — agreed, done

You were right to push on this one. Split in two: `SaxoException` and `IndexError` stay at `warning` (routine short history), everything else logs at `error` with `exc_info=True`. A `TypeError` from a future detector change now surfaces with a traceback instead of blending into the routine lines while the scan reports success.

### `combo` raising for 60 ≤ len < 235 — closed, but not by faking the criterion

Real gap, and thanks for catching that deferring `macd0lag` only helped the *rejected* sets. `combo` now declines up front below `COMBO_MIN_CANDLES`.

I did not take the suggested route of making the `macd` criterion `False` on short history. That would emit a COMBO alert whose `macd` flag was never computed — a fabricated indicator value reaching the trading journal and the Slack digest, where `False` is indistinguishable from a genuine bearish crossover. It could also flip a signal's strength, since `macd` is one of the five criteria feeding the `>= 4` STRONG threshold. Declining follows the precedent already set for `mm50_touch`, which returns `None` below 60 rather than raising, and for the documented reason of not polluting logs with legitimately short histories (spec 019 research §"Insufficient history").

Worth flagging one side effect beyond the alerting path: `ComboWorkflow.init_workflow` calls `combo` with no handler, so a workflow on a short history used to raise into the engine and now gets `indicator_value = None`, i.e. no signal. That's the better behaviour, but it is a behaviour change outside this PR's stated scope — say the word if you'd rather it stayed loud there.

**`needed 104`:** fixed. `macd0lag` now derives its minimum from the periods (`signal_period * 9 + long_term_period * 6 - 2`) and checks it once up front; the per-step guard that reported `long_term_period * 4` is gone. Verified the empirical minimum is still exactly 235, so the formula and reality agree.

### Nits

- `233` → `COMBO_MIN_CANDLES` in the docstring: fixed.
- `test_short_history_never_raises` being vacuous: correct, and the same blind spot bit my own probe earlier. Added `test_a_scored_direction_on_short_history_returns_none`, which truncates a fixture that produces a real signal at full length, plus `test_the_minimum_matches_what_macd0lag_actually_needs` tying the constant to what `macd0lag` derives so they can't drift.
- **Eager `bb20`/`bb20_previous`/ATR in `_ComboContext`: not doing this one.** You called it cosmetic and I agree — measured, they're ~0.05ms against a rejection path already down to 0.4ms, so lazy properties would add boilerplate for no measurable gain. Happy to revisit if the rejection path ever shows up in a profile.

Re-ran the differential after these changes: 503 windows of ≥235 candles, still zero divergence. The five now-improved cases are all short fixtures (12–61 candles) that previously raised.

---

RNiveau:
## Re-review after 212fdcc + d6f1023

All four points are addressed, and CI is green (backend + frontend). I re-ran the same verification against the new head.

**Behaviour, re-checked.** Same differential harness (every `*.obj` fixture, one-candle sliding window, 3484 windows, comparing the full signal including `details`):

```
identical: 503   (67 signals, unchanged from before)
old raised / new returns None: 2981   (was 2800)
real divergences: 0
```

The +181 are windows where `main` **and** the previous head both raised, and which the new up-front guard now declines cleanly — that is the fix showing up, not new drift. I specifically checked the risk the `COMBO_MIN_CANDLES` guard introduces, i.e. whether it silently kills signals `main` used to produce on shorter histories:

```
main signals: 67   shortest window with a main signal: 235   main signals under 235 candles: 0
```

So nothing is lost: `main` never once produced a combo below 235 candles, it only raised. Moving the guard to the top of `combo` is strictly better than raising mid-scoring, and it makes the deferral purely a performance concern, which is easier to reason about.

Full suite: **829 passed, 10 skipped**. `mypy --explicit-package-bases --ignore-missing-imports --exclude pulumi .` → `Success: no issues found in 203 source files`. `flake8` and `black --check` clean.

The corrected `macd0lag` guard checks out too — `signal_period * 9 + long_term_period * 6 - 2` = 235 with the defaults, and it matches the derivation in the docstring (deepest read at offset 157 against an ema needing 78). Good catch replacing the per-step `long_term_period * 4` check, which reported less than half the real requirement; that misleading `needed 104` is what made the original minimum hard to see. `test_the_minimum_matches_what_macd0lag_actually_needs` is the right pin for the two numbers, and splitting `_safe_detect` so an unexpected exception logs at `error` with a traceback is exactly what I was after.

### One thing I would look at before merging

`_build_candles` requests `count=250` (plus at most one rebuilt intraday candle) against a minimum of 235. That is ~15 candles of headroom, and `combo` declining now logs at `debug` — so an instrument that comes back with fewer than 235 daily candles (a recent listing, a thin instrument, an API returning short) silently produces no COMBO alert, forever, with nothing visible in the Lambda logs. The old behaviour was wrong in the other direction (an exception that ate the asset's other alerts), so this is not a regression — but the failure mode went from loud-and-harmful to quiet-and-invisible. Either bump the fetch count to give real headroom, or log the decline at `info`/`warning` so a permanently combo-less asset is discoverable.

### Nit

`saxo_uic: Optional[str | int]` widens the production signature to fit the test. It is defensible — the value comes from `instrument.get("Identifier")`, which is an int, so `Optional[str]` was the annotation that was wrong — but then `Optional[int]` is the truthful fix, and the docstring at `alerting.py:98` still documents `saxo_uic: int`. `get_asset_detail` takes `int` while `get_historical_data` takes `str | int`, so the codebase is already inconsistent here; no need to settle it in this PR, just flagging that the union was chosen to satisfy a test rather than to describe a real caller.

Nothing blocking from me. LGTM.

---

PR #691, opened 2026-07-30
refactor: backtest strategy seam and combo model vocabulary (spec 026, phases 1-2)

RNiveau:
Phases 1–2 of [spec 026](https://github.com/RNiveau/saxo-order/blob/main/specs/026-combo-ger40-backtest/tasks.md) — the groundwork the combo engine needs. **No behavior change**: the six shipped backtests run exactly as before.

## Phase 1 — vocabulary (T001–T005)

- `ExitReason.END_OF_RUN`, and `Strategy.C5M` / `C15M` / `C1H`.
- `BacktestDefinition.unit_time` and `.combo_entry`, with `__post_init__` validation in the style the class already uses: a combo definition **must** carry a timeframe and `double_take_profit`, and **may not** carry any of the ten flags that describe the 9h reference range it doesn't have. Each would otherwise ship as a silent no-op.
- `Position.retarget(first_target, take_profit)` plus optional `h1_high`/`h1_low`. `structural_level` now raises `SaxoException` rather than inventing a level (constitution §II.5 — no `assert`).
- `ComboTwoLot`: the shared two-lot accounting with a `targets` that raises, because the mm20 and the opposite band only exist relative to a candle.

`retarget` is the whole reason the combo strategy needs no new exit policy — `DoubleTarget` already does TP1 → bank → arm break-even → TP2, and now works unchanged against levels that move.

## Phase 2 — the strategy seam (T006–T009)

The day-scoped engine moves **verbatim** into `SessionRangeStrategy`; `BacktestService` is reduced to registry + dispatch. A definition names the engine it runs on instead of the day loop carrying a fifteenth variant flag.

Selecting a combo definition **raises** until its engine lands in phase 3. Falling back to the session-range engine would report results for rules it never applied — the exact failure this seam exists to prevent.

## Verification

`test_backtest_golden.py` — which runs every registered definition against a fixed synthetic market and diffs a committed snapshot — **passes unregenerated**. That is the claim that the move is a move.

```
849 passed, 10 skipped     # full suite
flake8: 0 · mypy: clean (4 pre-existing third-party stub warnings) · black/isort clean
```

## One thing reviewers should look at

`tests/api/services/backtest/test_run_range.py` has six patch sites moved from `service.evaluate_day` to `service.strategies.session_range.evaluate_day`. Those tests patched `evaluate_day` on the service and called `run_range` on the same object — which only worked while both lived together. The delegation moved that call inside the strategy, so the patch stopped intercepting and the real code ran against a `MagicMock` candle service.

Production behavior is unchanged; these tests assert the internal call graph, which is precisely what `test_backtest_golden.py`'s docstring says it deliberately does not do, so the golden suite stayed green throughout.

Worth flagging: these six tests do not currently run in a fresh container — `pytest-mock` is declared in `pyproject.toml` but was not installed, so they error at setup and a plain `pytest` run reports them as errors alongside a green count. I installed it locally to catch this; otherwise the moved patch sites would have shipped silently broken.

## Next

Phase 3 (T010–T028) is the MVP: the candle source, the signal→entry state machine, the band levels and the engine itself.

---
_Generated by [Claude Code](https://claude.ai/code/session_0141Ck7BANUxeAZC51V6WnDQ)_


RNiveau:
## Review — spec 026 phases 1–2

### What it does

Two things, cleanly separated:

- **Phase 1 (vocabulary)**: `ExitReason.END_OF_RUN`, `Strategy.C5M/C15M/C1H`, `BacktestDefinition.unit_time` + `.combo_entry` with `__post_init__` rejection of the ten session-range flags, `Position.h1_high/h1_low` made optional with `structural_level` raising, `Position.retarget()`, and `ComboTwoLot`.
- **Phase 2 (seam)**: the day engine moves verbatim into `SessionRangeStrategy`, `BacktestService` becomes registry + dispatch through `StrategySelector`, and a combo definition raises rather than silently running on the wrong engine.

The move really is a move — I diffed `session_range.py` against the deleted `service.py` block and the bodies are identical (`_evaluate_from_candles`, `_evaluate_trades`, `_open_position`, `_take_profit_level`, `_reset`, `_fetch_daily_candles`), with only the `self.candle_source`/`self.logger` construction relocated. Combined with an unregenerated `test_backtest_golden.py`, the no-behavior-change claim holds.

The design calls are good ones. `retarget` genuinely does let `DoubleTarget` work unchanged against moving levels, and refusing to fall back to the session-range engine is the right failure mode for a half-landed strategy.

---

### Issues

**1. `build_lot_model` raises on a valid combo definition — `ComboTwoLot` is unreachable (`api/services/backtest/rules.py:122`)**

This is the one thing I'd fix before phase 3, and it's a live inconsistency introduced by phase 1 rather than a phase-3 gap. A combo definition is now valid with `double_take_profit=True`, `first_target_fraction=None`, `runner_extension_points=None` — exactly the combination `build_lot_model` rejects:

```python
if not definition.double_take_profit:
    return SINGLE_LOT
if definition.runner_extension_points is not None:
    return ExtendedTwoLot(...)
if definition.first_target_fraction is None:
    raise ValueError("double_take_profit requires first_target_fraction or ...")
```

And `build_exit_chain` calls it unconditionally (`two_lot = isinstance(build_lot_model(definition), TwoLotAccounting)`), so the *first* thing the combo engine does in phase 3 will explode. Its docstring is also now factually wrong:

> BacktestDefinition already rejects a double take-profit with no first-target fraction

It doesn't anymore — this PR added the `not self.combo_entry` exemption in `model/backtest.py`. Two answers to one question, which is what the docstring itself says the refactor removed. Suggest adding the branch now, since `ComboTwoLot` exists and has tests but nothing can reach it:

```python
if definition.combo_entry:
    return ComboTwoLot()
```

**2. No test covers `StrategySelector` — the phase-2 deliverable**

`grep` over `tests/` finds no reference to `StrategySelector` or `for_definition`. The two behaviors the seam exists for are both untested:

- `for_definition(get_definition("G9H")) is selector.session_range`
- a combo definition raises `SaxoException` (and does *not* return the session-range engine)

The second matters most: it's the guarantee the PR body leads with, and no registered definition sets `combo_entry` yet, so the raise branch has never executed. `test_definitions.py` covers the model validation well — this is the layer above it that's uncovered.

**3. Name collision: `Strategy` protocol vs `Strategy` enum**

`api/services/backtest/strategy.py` defines `Strategy` (Protocol), while `model/enum.py` defines `Strategy` (the enum) — and `definitions.py` imports the enum by that name in the same package. Nothing breaks today, but phase 3's engine module will plausibly want both. `StrategyEngine` or `BacktestStrategy` for the protocol would cost nothing now and avoids an aliased import later.

---

### Smaller things

- **`StrategySelector` docstring describes an unwritten future.** "Both engines need the same collaborators, so they are constructed here" — there is one engine. The `dynamodb_client` parameter is accepted and dropped. Keeping the signature is fine (phase 3 will use it); the docstring reads as if the second engine already exists.
- **Dead attributes.** `SessionRangeStrategy.dynamodb_client` is assigned and never read (only forwarded to `CandleSource`), and `BacktestService.dynamodb_client` is now unused entirely. `BacktestService.candles_service` is still read — but only by tests (`test_run_range.py:229`, `test_engine_impulsive.py:298`), which reach through it into the shared mock. Worth a note that it's a test handle now, not a collaborator.
- **`Position` docstring is stale.** "A position open on one side of the H1 range" — no longer true for a combo position, which is the whole reason `h1_high`/`h1_low` became optional.
- **`x or None` in `session_range_only`.** `self.structural_stop or None` and `self.stop_from_reference_level or None` work for booleans, but the idiom is a trap if either field ever becomes numeric (`0.0 or None` → `None`, check silently passes). The other eight entries are read directly, so an explicit `if self.structural_stop:` would be more uniform. Cosmetic.
- **Ordering of the combo checks is load-bearing and undocumented.** `test_a_combo_time_cut_is_rejected` already notes it (the older double-TP/time-cut rule fires first), but the same is true of `impulsive_close_fraction`: alone it hits the pairing rule, and only when paired does the combo block catch it. Fine as-is, just fragile if the block moves.
- **`retarget` accepts any level.** No guard that `take_profit` is on the favorable side of entry. The mm20 can sit *behind* price at signal time, in which case `DoubleTarget` fills TP1 on the first candle. Not a defect in this PR — flagging it as a phase-3 decision to make deliberately rather than discover from a golden diff.

---

### On the `pytest-mock` note in the PR body

CI is green on this PR (backend job, run `30582886632`), which means the six moved patch sites **did** run and pass there — `poetry install` picks up `pytest-mock ^3.15.1` and a missing plugin would have failed the job, not silently skipped it. So the moved patches are verified, not just locally.

The fresh-container failure is real though, and I reproduced it here (`ModuleNotFoundError: No module named 'pytest_mock'` under `poetry run`). The likely cause is `pyproject.toml:31` using the legacy `[tool.poetry.dev-dependencies]` section, deprecated since Poetry 1.2. Migrating it to `[tool.poetry.group.dev.dependencies]` would be a good standalone `chore:` — out of scope here, but worth an issue.

### Other checks

- **Security**: nothing — no I/O, credential, or input-handling surface touched.
- **Performance**: neutral. `CandleSource` is still built once per `BacktestService` (now one level down, inside the strategy), so no cache is dropped per request. The weekend skip and the single `_fetch_daily_candles` per `run_range` are preserved.
- **Conventions**: `SaxoException` over `assert`, existing enums used throughout, comments explain *why* rather than restating code. Consistent with `CLAUDE.md`.

**Verdict**: approve on the refactor — the seam is the right shape and the golden test earns the no-behavior-change claim. I'd land #1 (the `build_lot_model` combo branch, or at minimum the corrected docstring) and #2 (a `StrategySelector` test) before phase 3 builds on top of it.

---

RNiveau:
Thanks — #1 was a real defect and it had a sibling. Pushed `93a6166`.

## #1 — `build_lot_model` (fixed, and it was worse than one call site)

You're right on all counts: the `not self.combo_entry` exemption I added to `model/backtest.py` made a definition valid that `build_lot_model` rejects, so `ComboTwoLot` was unreachable and phase 3's first call would have raised. Added the branch before the placement fields are consulted, and rewrote the docstring — it asserted a rejection the same PR had removed.

**Writing the test for it surfaced the same bug a second time.** `build_exit_chain` appends `ArmBreakEven(params.break_even_trigger_points)` unconditionally, so a combo definition would have armed break-even on a favorable move of 20 points. FR-C08 says TP1 filling is the *only* path to break-even, and `break_even_trigger_points` is one of the three thresholds a combo run ignores entirely (FR-C16) — so it would have invented a rule from a number the spec says is meaningless there. Same failure mode as #1: shared code that doesn't know `combo_entry` exists. Fixed, and the chain is now asserted to be exactly `[Stop, DoubleTarget]`, with a companion test that session-range definitions still arm on a favorable move.

Both were invisible because no registered definition sets `combo_entry` yet — which is exactly your point in #2.

## #2 — `StrategySelector` tests (added)

New `tests/api/services/backtest/test_strategy_selector.py`, six tests. The two you named, plus: the engine is built once and reused across calls (it holds a candle source), and an explicit "never falls back" assertion separate from the raise, since routing to the wrong engine silently is the failure the seam exists to prevent. The routing test walks `list_definitions()` and keys on `combo_entry` rather than listing codes, following `test_only_the_impulsive_variants_left_the_cash_session`.

## #3 — name collision (renamed)

`Strategy` protocol → `BacktestStrategy`. Agreed it costs nothing now and phase 3's engine module will want both names.

## Smaller things

All taken except two:

- `StrategySelector` docstring rewritten — no longer describes a second engine that doesn't exist, and now says why `dynamodb_client` is threaded through unused (phase 3's combo engine reads its cache through it; taking it now keeps that change from touching every call site).
- `BacktestService.dynamodb_client` dropped. `candles_service` kept with a comment saying plainly that it's a test handle, not a collaborator.
- `Position` docstring updated.
- `x or None` replaced — but with plain truthiness (`if value:`) rather than `is not None`. Every entry in that dict treats `0`/`0.0` as unset, so truthiness is the correct predicate for all ten and the trap you described doesn't reappear if one becomes numeric.
- **Check ordering**: left as-is. Documenting each interaction in the block would go stale faster than the tests that pin them.
- **`retarget` accepts any level**: left deliberately. FR-C10 already handles it at the entry gate — an entry whose mm20 isn't strictly favorable is rejected outright, precisely so TP1 can't fire on the first candle and bank a loss. Guarding inside `retarget` would duplicate that rule somewhere that can't see the entry price at open time. Worth re-reading FR-C10 in phase 3 to confirm that's the right place.

## On `pytest-mock`

You're right and my PR note overstated it — CI green does mean those six patch sites ran and passed, so they were verified, not merely locally checked. The fresh-container failure is a container-setup issue, not a CI gap, and `[tool.poetry.dev-dependencies]` being the legacy section is a good lead. Happy to open the `chore:` issue for the migration if you want it tracked.

Full suite: 859 passed, 10 skipped. `test_backtest_golden.py` still passes unregenerated.

---

RNiveau:
## Second pass

Deeper look, mostly at things I took on trust the first time. Two items from the earlier comment still stand (`build_lot_model` raising on a valid combo definition; no test for `StrategySelector`) — not repeated here.

### The "verbatim move" claim: mechanically verified

I stopped eyeballing it and diffed the moved methods by AST — extracting each method source from `service.py@2bdcf4c` and from `session_range.py@f22cd5b` and comparing the segments:

```
IDENTICAL: _evaluate_from_candles
IDENTICAL: _evaluate_trades
IDENTICAL: _fetch_daily_candles
IDENTICAL: _open_position
IDENTICAL: _reset
IDENTICAL: _take_profit_level
IDENTICAL: evaluate_day
IDENTICAL: run_range
DIFFERS:   __init__
-        self.logger = Logger.get_logger("backtest_service")
+        self.logger = Logger.get_logger("backtest_session_range")
```

Byte-identical across all eight, one difference in the whole move. That is a stronger claim than the golden test alone makes, and it holds.

**The logger rename is the one real behavior delta.** Log lines from the engine now emit as `k-order.backtest_session_range` instead of `k-order.backtest_service` (`utils/logger.py` formats `%(name)s`). Nothing in the repo keys off the old name and `Logger.get_logger` takes no config per name, so this is safe — but it is worth knowing that any external log filter or CloudWatch metric filter matching `backtest_service` goes quiet. Intentional and correctly named; just not "no change" for anyone reading logs.

### New findings

**4. `unit_time` has no reverse validation (`model/backtest.py`)**

The PR validates *combo requires `unit_time`* but not the mirror. This is accepted today:

```python
BacktestDefinition(code="G9H", ..., unit_time=UnitTime.M15)   # combo_entry=False
```

Nothing reads `unit_time` outside the combo path, so it ships as a silent no-op — the exact failure mode the ten-flag rejection block exists to prevent, just pointed the other way. The block is thorough about combo definitions carrying session-range flags and silent about a session-range definition carrying a combo flag. Suggest:

```python
if self.unit_time is not None and not self.combo_entry:
    raise ValueError(
        "unit_time is only used by a combo_entry definition - a "
        f"session-range backtest is fixed at H1/5-minute (definition {self.code!r})"
    )
```

**5. `unit_time` isn't constrained to the timeframes the spec supports**

`UnitTime` carries `D`, `M5`, `M15`, `M30`, `H1`, `H4`, `W`, `M`. Spec 026 is 5m / 15m / H1, and `Strategy.C5M/C15M/C1H` say so — but `combo_entry=True, unit_time=UnitTime.W` registers without complaint. Since this whole block is about rejecting unhonorable combinations at registration, a membership check against `{M5, M15, H1}` belongs next to the `is None` check rather than being discovered in phase 3 by a weekly candle fetch.

**6. `StrategySelector`'s docstring is contradicted by the DI layer**

> Both engines need the same collaborators, so they are constructed here rather than per call: an engine holds a candle source, and rebuilding it per request would drop whatever it had already fetched.

`api/dependencies.py:140` — `get_backtest_service` is **not** `@lru_cache()`d, unlike `get_trade_republic_service` directly above it. FastAPI builds a fresh `BacktestService` per request, so the selector, the strategy and the `CandleSource` are all rebuilt per request regardless of what the selector does. And `CandleSource.__init__` stores three collaborator refs and nothing else — its cache is DynamoDB, there is no in-memory state to drop.

So the reasoning is wrong on both halves. The code is fine (pre-existing, no regression, and building the engines once per selector is still the right shape) — it's the new docstring asserting a caching property the system doesn't have. Either trim it to "constructed once so a request doesn't pay for two engines it won't use", or `@lru_cache()` the provider and make the claim true.

**7. `api/services/backtest/__init__.py` package docstring is stale**

Still describes the old layout:

> - `service` orchestration: fetch/cache, per-day evaluation, range runs

`service` is now registry + dispatch, and neither `strategy` nor `session_range` is listed — which is unfortunate given that this docstring is the map someone reads before looking for where evaluation lives, and phase 2's whole point is that it moved.

**8. `retarget` is safe only because of a validation rule two files away**

`Position.stop_level` returns `first_target_level` once `trailed_to_first_target` is set, so `retarget` can in principle move a *stop* that was supposed to be locked in — including backwards. It can't happen today: combo definitions reject `trail_to_first_target_points`, so `TrailToFirstTarget` is never in the chain, and `DoubleTarget` guards its TP1 branch with `not first_target_taken` so a post-fill retarget is inert. The safety is real but entirely implicit. One line on `retarget` naming that dependency would keep phase 3 from quietly relaxing the validation and reintroducing it.

**9. Minor: one entry in `session_range_only` is unreachable**

`"time_cut_minutes"` can never fire — a combo requires `double_take_profit` (validated immediately above), and `double_take_profit` + time cut is rejected earlier in `__post_init__`. `test_a_combo_time_cut_is_rejected` already documents this. Harmless as defense-in-depth, and it becomes live the day the double-TP/time-cut restriction is lifted; noting it only so nobody assumes the parametrized rejection list covers it.

### Still good

Nothing in this pass changed my read of the seam itself. The `Strategy` protocol boundary is the right one, `SessionRangeStrategy`'s docstring is an accurate description of the family it holds, and refusing to dispatch a combo definition rather than silently running it on the wrong engine remains the correct call. Findings 4–9 are polish and one genuine validation gap, none of them blocking the merge.

---

PR #692, opened 2026-07-30
feat: implement the GER40 combo backtests (spec 026, phase 3)

RNiveau:
Phase 3 of [spec 026](https://github.com/RNiveau/saxo-order/blob/main/specs/026-combo-ger40-backtest/spec.md) — the MVP. `C5M`, `C15M` and `C1H` are now selectable in the Backtest menu and run end to end on the seam #691 built.

## The strategy, as implemented

| Rule | Where |
|---|---|
| Entry from `combo` on the definition's timeframe, MEDIUM+STRONG only | `signals.py` |
| Triggered → fill at the signal candle's close; untriggered → a **one-candle** pending stop level | `signals.py` |
| Stop 50 pts beyond the **signal** candle's adverse extreme | `combo_strategy.py` → `Position.initial_stop_price` |
| TP1 = mm20, TP2 = opposite Bollinger band, **re-read every candle** | `bands.py` + `Position.retarget` |
| Break-even only when TP1 fills | reused `DoubleTarget` |
| One position, held across nights and weekends, `end_of_run` at the range edge | `combo_strategy.py` |

The exit chain is `[Stop, DoubleTarget]`. Everything else — the two-lot accounting, the gap-fill, the summary, the router, the CSV exports — is reused unchanged, which is what #691 was for.

New modules: `combo_candle_source.py` (one contiguous series per run, cache-first under its own key namespace, 250-candle warm-up), `bands.py`, `signals.py`, `combo_strategy.py`. 68 new tests.

### 1. FR-C10 declines a lot of entries, and there's a structural reason

On the synthetic golden market, **C1H produces zero trades**: 5 signals across the whole range, 2 entries, and both rejected because the mm20 sits *behind* the fill.

That's the rule working exactly as specified — but the cause is structural, not incidental. A combo buys a pullback towards the **ma50**; TP1 is the **mm20**. In a trending market the ma50 sits *above* the mm20, so the entry lands past TP1 and FR-C10 rejects it. The stronger the trend, the more often this fires — and a sloping ma50 is a precondition of the combo signal itself.

I have **not** changed the rule to work around this: it's what you specified, and the guard is sound in isolation (without it TP1 fills on the next candle and banks a loss as a take-profit). But it may mean the H1 timeframe is close to untradeable under this exit design. Worth running against real GER40 data before drawing conclusions, and it may argue for revisiting either TP1's placement or the entry-price convention.

C5M (59 trades) and C15M (22 trades) are unaffected on the same market.

### 2. The golden fixture had to learn about timeframes

It served a single H1 *reference* candle per day, so a combo definition saw ~21 candles where `combo` needs 235 — the strategy could not fire at all. It now serves a whole session at the requested timeframe.

**The discriminator is window width, not the timeframe.** My first attempt keyed on `ut`, which silently captured the session strategy's own 5-minute call and changed **all seven** existing snapshots. Caught by diffing the regenerated file per definition rather than trusting the test to pass. A combo asks for a 20-hour day; the widest a session backtest asks for is 12.

## Verification

```
942 passed, 10 skipped
flake8 0 · mypy clean · black/isort clean
```

`backtest_golden.json` was regenerated once (T028). I diffed it per definition: the seven existing entries are **byte-identical**, only `C5M`/`C15M`/`C1H` were added.

The `test_golden_market_actually_produces_trades` guard now carries two named exemption sets rather than silently skipping — `C1H` for the FR-C10 finding above, and all three combo codes for the detail-day check, since a position opened before the detail day and closed after it correctly leaves that day showing no trade under FR-C15's entry-day attribution.

## Not in this PR

Phase 4 (`tunable_parameters` so the UI stops offering three inert inputs) and phase 5 (day-detail view, CSV). Both are ergonomics — the strategy is measurable through the API today.

---
_Generated by [Claude Code](https://claude.ai/code/session_0141Ck7BANUxeAZC51V6WnDQ)_


RNiveau (commented):
Read the whole thing and ran the engine against the golden fixture to check a couple of claims. The decomposition is good — `Side`, `Position.retarget`, the policy chain and the two-lot accounting all carry over without a special case, and the exit chain really is just `[Stop, DoubleTarget]`. The cache namespace split and the window-width discriminator in the fixture are both the right call.

One blocking bug, and one place where the PR description's diagnosis doesn't match what the code actually does.

---

## 1. Trades entered during the warm-up are counted in the summary but not in the day rows

`_run` walks the **entire** series, and `series = warm_up + run` (`combo_candle_source.py:104`). The warm-up is 250 candles and `combo` needs 235, so roughly the last 15 warm-up candles can fire a signal and open a position **before `start_date`**.

`_day_summaries` filters days to `start_date <= day <= end_date` and attributes trades by entry day, so those trades are dropped from the table. But `build_summary` at `api/services/backtest/combo_strategy.py:136` is passed the *full* `trades` list. The two disagree.

Measured, golden fixture, 2026-03-23 → 2026-04-30:

| | trades | entered before start | summary `final_result` | sum of day rows |
|---|---|---|---|---|
| C5M | 31 | 2 | -2617.49 | -2767.39 |
| C15M | 14 | 1 | **+411.60** | **+344.58** |

`number_of_trades` disagrees with `sum(day.trade_count)` too (31 vs 29, 14 vs 13).

Two problems beyond the arithmetic:

- the run reports P&L from trades the user did not ask about — the requested window is `start_date`..`end_date`;
- the result depends on where the warm-up boundary happens to land, so bumping `WARM_UP_CANDLES` for an unrelated reason silently moves every headline number.

Simplest fix: skip the entry branch of `_run` while `candle_date(candle).date() < start_date` (the warm-up should feed the indicator, not trade). Filtering `trades` before `build_summary` also works but leaves a position that opened pre-start still running into the range, which I think you *do* want — so gating the entry is the cleaner of the two. Either way it needs a test; the golden snapshot happily records the current behaviour.

## 2. The FR-C10 explanation in the description doesn't hold up — and it isn't a C1H-only problem

You describe the cause as: *"a combo buys a pullback towards the ma50; TP1 is the mm20; in a trending market the ma50 sits above the mm20, so the entry lands past TP1."*

The two C1H declines don't look like that:

```
DECLINE LONG   entry=24915.27  mm20=24802.83  ma50=24892.0  gap=-112.4  (triggered)
DECLINE SHORT  entry=24760.88  mm20=24824.78  ma50=24779.4  gap= -63.9  (pending)
```

On the long, mm20 (24802) is *below* the ma50 (24892) — the opposite of the stated ordering — and the entry sits above both. The mechanism looks like the **entry-price convention**, not TP1's placement: a triggered entry fills at the signal candle's *close*, and a pending entry fills at a stop beyond the prior extreme. Both are, by construction, prices reached only after a decisive move in the trade direction. On a slow timeframe with big candles, that move alone is often enough to clear the mm20 — so the fill lands past TP1 before the position exists.

Which also means the framing "C5M and C15M are unaffected" isn't right. Same range, same fixture, full 2026-03-02 → 2026-04-30:

| | entries | declined by FR-C10 | accepted |
|---|---|---|---|
| C5M | 131 | 54 (41%) | 77 |
| C15M | 51 | 22 (43%) | 29 |
| C1H | 2 | 2 (100%) | 0 |

Four in ten entries are dropped on every timeframe; H1 is the same effect with too small a sample to survive it. That's a much stronger argument for revisiting the entry-price convention than the H1 zero is on its own, and it changes what "worth running against real GER40 data" is meant to settle. I'd rewrite that section of the description before this merges — the code is fine, the recorded diagnosis is what future-you will read.

---

## Smaller things

- **`_run` never evaluates the entry candle against the exit chain** (`combo_strategy.py:185` opens, then the loop `continue`s). Harmless for a triggered entry — the stop is measured from the same candle it fills on, at that candle's close. But a *pending* fill happens on a later candle while the stop still sits beyond the **signal** candle's adverse extreme, so that stop can be inside the filling candle's range and gets skipped. 0 occurrences on the synthetic market, so it's theoretical here, but it's an optimistic bias on real intraday data.

- **A pending entry survives nights and weekends.** `feed` resolves it against the next candle *in the series*, which after Friday 22:00 is Monday 02:00. The fill is conservative (`side.worse`), so it isn't optimistic — but is carrying a stop order over a weekend intended? FR-C04's "one candle" reads like it means one candle of live market.

- **`evaluate_day` is a different backtest from the range run** (`combo_strategy.py:85`). Fresh one-day run, own warm-up, `END_OF_RUN` at 22:00 — so the day view can show a trade the range run never took, and hide one it did. The docstring says so, which is good; when phase 5 ships the detail view, that caveat needs to reach the UI, not just the source.

- **`_day` caches the current, partial day permanently** (no TTL, no "not today" guard). Pre-existing — `CandleSource` does the same — but it costs more here, because a truncated day becomes permanent *warm-up* for every later run on that instrument, not just one day's result.

- `combo_candle_source.py:70`: `', '.join(str(u) for u in HORIZONS)` renders `UnitTime.M5, UnitTime.M15, ...`. `u.value` reads better in the message.

- `resolve_parameters` silently discards the three inert overrides for a combo definition. The comment argues against letting someone tune a number and see nothing move — but silently dropping it is that same failure, just moved. A 400 would be honest. Fine to leave if phase 4 is close.

Nothing else structural. Fix (1), reword (2), and I'm happy with this.

RNiveau:
Both landed in `689d72d`. Thanks — (1) was a real bug and (2) was me pattern-matching instead of reading the numbers.

## 1. Warm-up entries — fixed

Reproduced your figures exactly before touching anything (C15M +411.60 vs +344.58, 31 vs 29 trades). `_run` now takes the range start and gates the **entry** on it; the warm-up still feeds the search, since the first tradeable candle would otherwise have no history.

I took your suggested fix over filtering `trades` for the reason you gave — and the run's identity matters here: `start_date` is what the user asked about, so a trade opened before it isn't "excluded from the table", it shouldn't exist.

On your range, summary and day rows now agree exactly:

| | summary trades | day rows | summary result | day rows |
|---|---|---|---|---|
| C5M | 29 | 29 | -2767.3877 | -2767.3877 |
| C15M | 13 | 13 | 344.5791 | 344.5791 |

Three tests: no position opens before `start_date`, the first candle of the range *is* tradeable (it's a boundary, not a delay), and the warm-up still advances the search. C5M's snapshot moves 54 → 52; the other eleven are byte-identical.

## 2. The diagnosis — you're right, and it was worse than wrong

I asserted an ordering I never measured. Your decline log shows mm20 (24802) *below* the ma50 (24892) on the long — the opposite. Reproduced your table:

| | entries | declined | |
|---|---|---|---|
| C5M | 131 | 54 | 41% |
| C15M | 51 | 22 | 43% |
| C1H | 2 | 2 | 100% |

PR description rewritten around the entry-price convention, with the old claim marked as replaced rather than quietly deleted. You're right that it's the recorded diagnosis future-me reads — and "~40% of signals discarded on every timeframe" is a decision to take deliberately, against **FR-C03**, not the H1 curiosity I filed it as.

## Smaller things

**Fixed:**
- Partial current day no longer cached. Your point that it poisons the *warm-up* rather than one day's result is what made it worth doing now — `is_today_not_yet_closed` already existed.
- `u.value` in the horizon message.

**Deliberately not fixed, flagged instead:**

- **The entry candle skipping the exit chain.** Real, and I agree on the direction of the bias. But the correct fix depends on intra-candle ordering the data doesn't carry: for a *triggered* entry the fill is that candle's close, so running the chain over the candle's full range would stop the position out on a low that happened before it existed — a pessimistic bug replacing an optimistic one. Only the pending case is genuinely wrong, and separating them means the engine treating its two entry kinds differently. `SessionRangeStrategy` has the same `continue`, so this is a shared convention. Worth its own change with both engines in scope, not a quiet fix here.

- **Pending entries surviving weekends.** You've read FR-C04 the way I now think it was meant — "one candle" as one candle of live market. The spec doesn't say, and it's your rule. Happy to gate it on a same-session check if you confirm; I didn't want to reinterpret a requirement from a review comment.

- **`resolve_parameters` silently discarding.** Agreed the silent drop is the same failure moved. Phase 4 is next and removes the inputs at source; a 422 is the better answer if it slips.

- **`evaluate_day` being a different backtest.** Noted for phase 5 — the caveat needs to reach the UI, not just the docstring. Added to my phase 5 notes.

CI green. The earlier `backend` failure was on `0aa12bd`, before I refreshed the combo snapshots for #708's `bb_first` fix.

---

RNiveau (commented):
Re-checked at `689d72d`. The blocking bug is fixed and the description is corrected — this is good to merge from my side.

**Warm-up gate verified.** Re-ran the same measurement against the fixed head; summary and day rows now agree exactly, and nothing enters outside the window:

| range | | trades | Σ day rows | `final_result` | Σ day points | pre-start |
|---|---|---|---|---|---|---|
| 03-23 → 04-30 | C5M | 29 | 29 | -2767.3877 | -2767.3877 | 0 |
| | C15M | 13 | 13 | +344.5791 | +344.5791 | 0 |
| 03-02 → 04-30 | C5M | 52 | 52 | -1351.7190 | -1351.7190 | 0 |
| | C15M | 21 | 21 | +478.5946 | +478.5946 | 0 |

Gating the entry rather than filtering the trade list is the right half to gate — a position that legitimately opens on `start_date` still runs into the range, and the warm-up still feeds the search. The three tests cover the boundary in both directions, which is the part that would rot silently otherwise.

`1045 passed`, flake8/black clean, mypy clean apart from the pre-existing `aioboto3` stub warning.

**One latent thing in the gate, not a blocker.** `candle_time.date() < start_date` compares a *UTC* candle date against a *Paris* trading date. I checked both DST halves for `DaxCfdMarket` — 01:00→21:00 UTC in winter, 00:00→20:00 in summer, same UTC date as the Paris date either way — so it's correct for every definition that exists today, with an hour of margin. It stops being correct for a market opening before 01:00 Paris in winter, whose session would start on the previous UTC date and lose its first day here. That's shared with `paris_session_start_utc`, which builds the UTC instant from the Paris date's y/m/d, so it's a pre-existing constraint rather than something this PR introduces. Worth a comment at the constraint, not a fix.

**Also good:** `is_today_not_yet_closed` was the right existing helper for the partial-day cache — reusing it means the two candle sources can't drift on what "closed" means.

**Still open from the first pass, all as I filed them** — nothing here blocks:

- the entry candle is never run through the exit chain, so a pending fill can skip a stop that sits inside its own filling candle (0 occurrences on the synthetic market; an optimistic bias on real data);
- a pending entry survives nights and weekends — Friday 22:00 resolves against Monday 02:00;
- `evaluate_day` remains a structurally different backtest from `run_range`; that caveat needs to reach the UI when phase 5 lands;
- `resolve_parameters` still drops the three inert overrides silently — fine if phase 4 is close.

The FR-C10 rewrite is better than what I asked for: naming the entry convention (FR-C03) as the thing to revisit rather than TP1, and keeping the superseded claim visible instead of quietly replacing it, is the version that will still be readable in six months.

RNiveau:
Comment added at the site in `1ae22ac`, recording exactly the analysis you did — UTC candle date against Paris trading date, valid for every current market in both DST halves with an hour of margin (`DaxCfdMarket`'s 02:00 Paris open is 01:00 UTC in winter, 00:00 in summer), breaking for a market opening before 01:00 Paris in winter, and shared with `paris_session_start_utc` since that builds its instant from the Paris date's y/m/d.

Agreed it wants a comment rather than a fix: no market violates it, and the honest fix is converting the candle to Paris local before comparing, which would be the same latent assumption moved one layer down unless `paris_session_start_utc` changes with it.

The four still-open items stand as you filed them. On the pending-entry-over-weekend one I've since measured what it actually costs, and it's a bit sharper than "is this intended": the fill is conservative, but only the *entry* moves with a gap while FR-C06 anchors the stop to the signal candle, so risk inflates silently — 150 → 180 pts on a small gap. FR-C10 happens to decline the extreme cases (a big gap puts the entry above the lagging mm20), so what survives is a 10–30% understatement of risk rather than a blow-up. 0 occurrences on the synthetic market; routine on real 5m data across a 22:00→02:00 boundary. I've asked which way you want it — one condition in `ComboEntrySearch.feed` if you want pending levels confined to a session.

CI green at `1ae22ac`.

---

PR #693, opened 2026-08-01
feat: add mm7 break detection to the alerting system

RNiveau:
Breaking the 7-period moving average flags a short-term change of
character, which the daily scan had no detector for.

A raw MA7 cross fires on a large share of the universe every day, so the
detector qualifies it: the close has to clear the average by 0.5% and the
previous 3 candles must all have closed on the other side. That keeps
chop around a flat average out of the digest, the same way mm50_touch
gates on proximity and slope.

The triage prompt gets the pattern semantics it needs to weigh it: a
short-term timing trigger, not a thesis, read against the sign of
ma50_slope, and never enough on its own for a "high" conviction.

Claude-Session: https://claude.ai/code/session_01RgW8qZTKpHXDcuc9H5yQLK

### DIFF
--- a/services/indicator_service.py
+++ b/services/indicator_service.py
@@ -148,6 +148,13 @@ def mm50_touch(candles: List[Candle]) -> Optional[Dict[str, float]]:
 
 
 def _mm7_at(candles: List[Candle], offset: int) -> float:
+    """
+    The MM7 as it stood at candles[offset], i.e. that candle and the 6 older
+    ones. Newest is index 0, so dropping the first `offset` entries walks
+    backwards in time and `mobile_average` then reads the 7 newest of what
+    remains - candles[offset:offset+7]. Same idiom as mm50_touch's
+    mobile_average(candles[10:], 50) for the MA50 of 10 candles ago.
+    """
     return mobile_average(candles[offset:], MM7_PERIOD)
 
 
--- a/services/indicator_service.py
+++ b/services/indicator_service.py
@@ -1,5 +1,5 @@
 import datetime
-from typing import Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import numpy
 
@@ -103,6 +103,16 @@ def mobile_average(candles: List[Candle], period: int) -> float:
 MM50_TOUCH_PROXIMITY = 0.01
 MM50_TOUCH_SLOPE_MIN = 3.0
 
+MM7_PERIOD = 7
+# A close that merely grazes the average is not a break: price has to clear it
+# by a margin to count.
+MM7_BREAK_MIN_DISTANCE = 0.005
+# ...and it has to flip a short-term run rather than cross back and forth. A
+# raw cross fires on a large share of the universe every day, which would
+# drown the daily digest instead of informing it.
+MM7_BREAK_MIN_STREAK = 3
+MM7_BREAK_MIN_CANDLES = MM7_PERIOD + MM7_BREAK_MIN_STREAK
+
 COMBO_MA50_SLOPE_MIN = 3.0
 COMBO_MA50_SLOPE_STRONG = 10.0
 COMBO_BB_FLAT_SLOPE_MAX = 5.0
@@ -134,6 +144,68 @@ def mm50_touch(candles: List[Candle]) -> Optional[Dict[str, float]]:
         "ma50": ma50_last,
         "distance_pct": (close - ma50_last) / ma50_last * 100,
         "slope": slope,
+    }
+
+
+def _mm7_at(candles: List[Candle], offset: int) -> float:
+    return mobile_average(candles[offset:], MM7_PERIOD)
+
+
+def _mm7_streak(candles: List[Candle], direction: Direction) -> int:
+    """
+    Number of consecutive candles before the break that closed on the side the
+    price is leaving. Stops as soon as the history runs short of a full MM7.
+    """
+    streak = 0
+    offset = 1
+    while len(candles) - offset >= MM7_PERIOD:
+        mm7 = _mm7_at(candles, offset)
+        close = candles[offset].close
+        on_previous_side = (
+            close >= mm7 if direction == Direction.SELL else close <= mm7
+        )
+        if not on_previous_side:
+            break
+        streak += 1
+        offset += 1
+    return streak
+
+
+def mm7_break(candles: List[Candle]) -> Optional[Dict[str, Any]]:
+    """
+    Detect the last candle closing through the 7-period mobile average.
+
+    A break is a short-term timing signal, not a thesis: it only fires when
+    the close clears the average by MM7_BREAK_MIN_DISTANCE and the previous
+    MM7_BREAK_MIN_STREAK candles all closed on the other side, so chop around
+    a flat average stays silent.
+    """
+    if len(candles) < MM7_BREAK_MIN_CANDLES:
+        return None
+
+    mm7 = _mm7_at(candles, 0)
+    close = candles[0].close
+    distance = (close - mm7) / mm7
+
+    if distance < -MM7_BREAK_MIN_DISTANCE:
+        direction = Direction.SELL
+    elif distance > MM7_BREAK_MIN_DISTANCE:
+        direction = Direction.BUY
+    else:
+        return None
+
+    streak = _mm7_streak(candles, direction)
+    if streak < MM7_BREAK_MIN_STREAK:
+        return None
+
+    return {
+        "close": close,
+        "mm7": mm7,
+        "previous_close": candles[1].close,
+        "previous_mm7": _mm7_at(candles, 1),
+        "distance_pct": distance * 100,
+        "direction": direction.value,
+        "streak": streak,
     }
 
 
--- a/frontend/src/components/AlertCard.tsx
+++ b/frontend/src/components/AlertCard.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import type { AlertItem } from '../services/api';
 import { getTradingViewUrl } from '../utils/tradingview';
+import { getAlertTypeLabel } from '../utils/alertLabels';
 import { assetDetailsService } from '../services/api';
 import './AlertCard.css';
 
@@ -26,13 +27,6 @@ export const AlertCard: React.FC<AlertCardProps> = ({ alert, onExclude }) => {
     });
   };
 
-  const formatAlertType = (type: string): string => {
-    return type
-      .split('_')
-      .map(word => word.charAt(0).toUpperCase() + word.slice(1))
-      .join(' ');
-  };
-
   const getAgeLabel = (hours: number): string => {
     if (hours < 1) return 'Just now';
     if (hours === 1) return '1 hour ago';
@@ -42,7 +36,7 @@ export const AlertCard: React.FC<AlertCardProps> = ({ alert, onExclude }) => {
     return `${days} days ago`;
   };
 
-  const formatMA50Slope = (slope: unknown): string => {
+  const formatSignedPercent = (slope: unknown): string => {
     // Handle null, undefined, or non-numeric values
     if (slope === null || slope === undefined || slope === '') {
       return 'N/A';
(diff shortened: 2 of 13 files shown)


RNiveau:
## Review: `feat: add mm7 break detection to the alerting system`

### Overview

Adds `AlertType.MM7_BREAK`: a pure detector in `services/indicator_service.py` that fires when the latest close clears its 7-period MA by >0.5% after ≥3 prior candles closed on the other side, wired into `run_detection_for_asset` exactly like `mm50_touch`, plus a pattern-semantics entry in `TRIAGE_SYSTEM_PROMPT` and a frontend label. Backend is ~90 lines of code; the rest is `specs/027-*` docs and tests.

Verified locally on the head commit: `pytest` on the three touched test modules → **138 passed**; `mypy`, `flake8`, `black --check` clean.

### What's right

- Candle ordering respected throughout — `_mm7_at(candles, i) = mobile_average(candles[i:], 7)` correctly evaluates each candle against the MA it could actually have known (plan §Key Design Decisions 2), not against today's value.
- `Direction.BUY`/`SELL` reused rather than an ad-hoc `"up"`/`"down"` string, per CLAUDE.md.
- `MM7_BREAK_MIN_CANDLES` derived from `MM7_PERIOD + MM7_BREAK_MIN_STREAK` instead of a literal `10` — re-tuning the streak can't silently desync the guard.
- `_mm7_streak` stops on short history instead of letting `mobile_average` raise; the `detect(...)` wrapper is used so a failing detector doesn't cost the others their `store_alerts` call. No extra Saxo calls.

### Issues

**1. The deterministic triage fallback does change behavior — T008 only checked half the case.** (medium)

`_fallback_conviction` returns `HIGH` at `family_count >= 2`. `MM7_BREAK` is its own family, so FR-010 holds for an MM7 break *alone* (1 family → `WATCH`) — that part is correct. But any asset that previously had exactly one family and now also fires `mm7_break` gets promoted **`WATCH` → `HIGH`** on the fallback path. Given the whole premise of the PR is that MA7 crosses are common, that's a systematic inflation of the fallback `HIGH` tier, not an edge case. `tasks.md` T008 records "no change required" without examining the co-firing case, and the Risks table doesn't mention it.

Worth either excluding `MM7_BREAK` from the family count in `_fallback_conviction` (a timing trigger shouldn't be the pattern that promotes to `HIGH`, which is exactly what the new prompt entry tells the *LLM* path) or acknowledging the shift explicitly.

**2. The frontend label map edit doesn't reach the alert badge.** (low)

`ALERT_TYPE_LABELS` in `Alerts.tsx` feeds only the filter dropdown (line 180). The card badge goes through a *separate* `formatAlertType` local to `AlertCard.tsx:29`, which title-cases the raw value — so the UI shows **"MM7 Break"** in the filter and **"Mm7 Break"** on the card. Two labelling paths that now disagree.

This also makes the `tasks.md` T009 note inaccurate: without the map entry `mm50_touch` rendered as raw text in the *dropdown*, not everywhere. And `double_bottom` is still absent from the map — same drive-by class of bug the PR fixed for `mm50_touch`. Suggest exporting one `getAlertTypeLabel` helper and having `AlertCard` use it.

**3. SC-005 isn't met by the UI.** (low)

"A reader of the Slack digest or the alerts UI can tell at a glance which direction the break went and how far price cleared the average" — `AlertCard` surfaces only `ma50_slope` (line 102); `direction` and `distance_pct` land in the collapsed `JSON.stringify(alert.data)` dump at line 181. Not "at a glance". The Slack digest is triage prose so it likely covers it, but the UI half of SC-005 is unfulfilled. Either render `direction`/`distance_pct` on the card or relax the criterion.

**4. Test gaps at exactly the boundaries the spec calls out.** (low)

The spec makes the threshold semantics an explicit assumption ("distance exclusive, streak inclusive"), but neither boundary is pinned:

- No test at exactly 0.5%. I checked — it correctly returns `None` — but nothing stops a future `<=` from flipping it silently.
- No test for `streak == 2 → None`. The chop test yields streak 1, so `MM7_BREAK_MIN_STREAK` is never exercised from below. Constructing a genuine streak-2 series is less obvious than it looks (in a rising series every prior close sits above its own trailing MA, so streaks run long) — which is precisely why the test earns its keep.
- Both happy-path tests (`[95.0] + [100.0]*9`, `[105.0] + [100.0]*9`) satisfy the streak *only* through the equality branch: every prior close is exactly equal to its own MM7. The inclusive-touch rule is well covered; the ordinary strictly-one-sided run isn't tested at all.

**5. `distance_pct` is measured against an MA that includes the breaking candle.** (nit)

The breaking close carries 1/7 weight in `mm7`, dragging the average toward itself, so a large move's measured distance is systematically smaller than its distance from the pre-break average. The 0.5% gate is therefore a bit stricter than the spec's phrasing ("clear the average by 0.5%") reads. Consistent with how `mm50_touch` does it, so probably fine as-is — but it's a real factor when T010 re-tunes the threshold, and the payload already carries `previous_mm7` if you'd rather measure against that.

### Verdict

Detector logic is correct and the wiring is faithful to the `mm50_touch` precedent. Item 1 is the one I'd want resolved before merge — it's an unintended behavior change to the fallback digest, in the opposite direction from FR-010's intent. Items 2–5 are polish.

---

RNiveau:
Thanks — item 1 was a real bug and my T008 note was the cause of it. Addressed 1–4 in `5da5cc3`; 5 is acknowledged as a documented tradeoff, details below.

### 1. Fallback promotion — fixed

You're right, and the error was that T008 stopped once the lone-MM7 case satisfied FR-010. The co-firing case is the one that matters, and `>= 2` promoted it.

`_TIMING_PATTERNS = {AlertType.MM7_BREAK}` is now excluded from the confluence count via `_structural_families()`, used by `_fallback_conviction`, `_fallback_rationale`, and the ranking sort.

One thing your suggestion doesn't mention that the fix needed: dropping MM7 from the count alone would have broken the *other* direction. The `watch` gate tested `family_count == 1`, so a co-firing asset would have gone `2 → not high`, then failed the `== 1` test and landed in `noise` — worse than the promotion it was fixing. The gate is now `structural_count >= 1`, so such an asset keeps exactly the tier it had before this feature existed. That equivalence is what `test_fallback_does_not_let_mm7_break_promote_to_high` pins.

A lone MM7 break is now `noise` on the fallback path. That's deliberate: with reasoning unavailable, a bare timing trigger isn't actionable. The alert is still stored and still shows in the UI — only the fallback ranking ignores it. Recorded in spec Assumptions and as **FR-013**.

### 2. Two labelling paths — fixed

Confirmed exactly as described: `ALERT_TYPE_LABELS` fed the dropdown at `Alerts.tsx:180`, `AlertCard.tsx` title-cased the raw value. Extracted to `frontend/src/utils/alertLabels.ts` with one exported `getAlertTypeLabel`, used by both. It falls back to title-casing for unmapped types, so an unmapped type degrades to `Mm7 Break` rather than raw `mm7_break` — no alert type can regress to snake_case again. `double_bottom` added. The inaccurate T009 note is corrected in `tasks.md` rather than quietly edited away.

### 3. SC-005 — rendered rather than relaxed

The criterion was mine, so I'd rather make it true than weaken it. `AlertCard` now has a Direction row beside the MA50 slope — green `Buy` / red `Sell`, signed distance next to it — instead of leaving both fields in the collapsed JSON. It keys off `data.direction`, so `combo` alerts get it too.

### 4. Boundary tests — added

Four tests. On the streak-2 case you were right that it's fiddlier than it looks — a plain rising series gives long streaks, so it needed a deliberate dip at index 3 (`[90, 101, 101, 99, 100…]`) to produce a genuine 2-candle run. Also added: exactly 0.5% → `None`, with a companion firing just past it on the same shape so the distance gate is isolated from the streak rather than both being asserted by one test; and a strictly one-sided run, since both original happy paths only exercised the equality branch.

### 5. `distance_pct` includes the breaking candle — acknowledged, not changed

Correct, and it does make the gate slightly stricter than the prose implies. Keeping it consistent with `mm50_touch` rather than having two proximity conventions in one file; `previous_mm7` is in the payload for anyone who wants the other measure. Now written down in spec Assumptions so T010's re-tuning doesn't rediscover it the hard way.

Local verification on `5da5cc3`: **873 passed, 10 skipped**; `black`/`isort`/`flake8` clean, `mypy` clean apart from the pre-existing `aioboto3` stub error; frontend `tsc --noEmit` clean, `eslint` 0 errors (3 pre-existing warnings in untouched files), `vite build` OK.

---

RNiveau:
Verified `5da5cc3` independently. All five items check out. One cosmetic leftover below, non-blocking.

### 1. Fallback promotion — confirmed fixed, and the equivalence claim holds exactly

I didn't take the "keeps exactly the tier it had before this feature existed" claim on trust — I enumerated every pattern combination up to size 3 across all `AlertType` members, at slopes `{0.0, 0.5, 1.0, 2.0, -5.0}`, and compared three formulas: pre-feature (`main`), the buggy intermediate (`6746dbd`), and `5da5cc3`.

- **vs. the buggy intermediate**: 38 tier changes, **all 38 involve `mm7_break`** — zero changes for any pattern set without it. So no collateral effect on existing behavior.
- **vs. pre-feature `main`**: **0 mismatches**. Every asset that appeared in the digest before this feature lands in the identical tier now.
- Every change is a de-escalation (`high→watch` 18, `high→noise` 18, `watch→noise` 2). Nothing escalates, which is the property that matters here.

Your point about the `== 1` gate is right and I'd missed it. Worth spelling out why `>= 1` is safe rather than merely convenient: `structural_count >= 2` returns early, so the second gate is only ever reached with a count of 0 or 1 — `>= 1` and `== 1` are provably identical there. It's a widening that cannot widen. The `congestion20 + mm7_break` case at slope 0 is the clearest illustration: `high → noise` looks alarming in isolation, but pre-feature that asset was `noise`, so the "drop" is just the phantom promotion being undone.

The 5 lone-`mm7_break` combinations are the only genuinely new digest entries — assets that wouldn't have appeared at all before. They come in as `noise` with `rank=None`, so they don't compete for attention. Reasonable, and FR-013 documents it.

### 2–4 — confirmed

- `getAlertTypeLabel` is now the single path; both `Alerts.tsx:180` and the `AlertCard` badge call it, and the `titleCase` fallback means an unmapped type degrades to `Mm7 Break` rather than raw snake_case. `double_bottom` present. `ma50-slope-neutral` exists in `AlertCard.css:153`, so `getDirectionClass`'s fallback branch isn't dead.
- Direction row renders off `data.direction`, so `combo` picks it up as you noted.
- The four new detector tests do isolate what they claim. I recomputed each: the 0.5% test lands on `mm7 == 100.0` exactly with `distance == -0.005`, and its companion at `99.4` fires on the same shape with the streak satisfied at 3 — so the distance gate really is pinned alone rather than incidentally passing because the streak failed. The streak-2 case genuinely produces `streak == 2` (candle 3's close of 99.0 falls under its own MM7 of ~99.857). The strictly-one-sided test has all three prior closes strictly above their own MM7 (106>103, 105>102, 104>101), covering the branch the original happy paths missed.

### Independent local run on `5da5cc3`

`pytest`: **873 passed, 10 skipped** — matches your number. `black` 213 files unchanged, `isort` clean, `flake8` exit 0, `mypy` 4 errors — all `import-untyped` stub misses (`aioboto3` ×3, `binance.error`), and `main` reports the same count, so pre-existing as stated. Frontend: `tsc --noEmit` clean, `eslint` 0 errors / 3 warnings (`AvailableFunds.tsx`, `AssetDetail.tsx`, `SearchResults.tsx` — none touched here), `vite build` OK.

### One leftover (nit, non-blocking)

`_fallback_rationale` now counts structural families but still lists every pattern, so a co-firing asset reads:

```
1 pattern(s): congestion20, mm7_break, slope 3.0%
```

Two patterns named, count of 1. Listing `mm7_break` is right — it's information the reader wants, and `test_fallback_does_not_let_mm7_break_promote_to_high` asserts its presence — but the count now measures something different from the list beside it. Something like `1 structural pattern(s)` or moving the timing trigger to its own clause would remove the apparent contradiction.

No objection to merging from my side once that's either addressed or waved through.

---

services/indicator_service.py
```diff
@@ -137,6 +147,68 @@ def mm50_touch(candles: List[Candle]) -> Optional[Dict[str, float]]:
     }
 
 
+def _mm7_at(candles: List[Candle], offset: int) -> float:
+    return mobile_average(candles[offset:], MM7_PERIOD)
+
```
RNiveau, line 159: Are you sure here ? First candle is the newest one

RNiveau: Yes — checked it rather than reasoned about it. Newest-first is exactly why the slice reads the way it does: dropping the first `offset` entries walks *backwards* in time, then `mobile_average` takes `[:period]` of the remainder, so the window is `candles[offset:offset+7]` — the candle at `offset` plus the 6 older ones.

Ran it on a series with closes 10 (newest) down to 1 (oldest), dates descending with index:

```
index 0  2026-01-10  close 10  (newest)
index 9  2026-01-01  close  1  (oldest)

_mm7_at(off=0) = 7.0  window [10, 9, 8, 7, 6, 5, 4]  2026-01-10..2026-01-04
_mm7_at(off=1) = 6.0  window [ 9, 8, 7, 6, 5, 4, 3]  2026-01-09..2026-01-03
_mm7_at(off=3) = 4.0  window [ 7, 6, 5, 4, 3, 2, 1]  2026-01-07..2026-01-01
```

Each window ends on its own candle and extends into the past — never into candles that hadn't happened yet. Same idiom as `mm50_touch`'s `mobile_average(candles[10:], 50)` for the MA50 of 10 candles ago.

Getting this backwards would have been the serious version of the bug — the streak check would compare each prior close against an average computed from candles *after* it, which is lookahead. It's what `test_fires_on_a_strictly_one_sided_run` guards: on a descending series, comparing against a forward-looking window would flip every comparison and the test would fail.

Fair question though, and the second time the slicing has needed explaining, so I've added a docstring on `_mm7_at` spelling out the direction (`bce8651`). No behavior change.

---

PR #696, opened 2026-08-01
feat: workflow-trigger collector for the alert triage agent (spec 028, phases 1-2)

RNiveau:
Phases 1–2 of [spec 028](https://github.com/RNiveau/saxo-order/tree/main/specs/028-triage-workflow-trigger). Foundational only: this adds the domain model and the collector that matches a day's workflow orders to the alert assets they corroborate. **Nothing consumes the result yet** — `TriageAgent` is untouched, so the daily brief is byte-for-byte what it was before.

## What's here

- **`WorkflowTrigger`** dataclass in `model/__init__.py`; `TriagedAsset` gains `workflow_triggers`, defaulting to `[]`.
- **`services/workflow_trigger_service.py`** — `collect_todays_triggers(dynamodb_client, run_date, alerts)`:
  - filters `workflow_orders` to the Paris session window, derived from `run_date` rather than the 18:15 schedule, so manual runs compute their own window;
  - matches each order to a scanned asset by `order_code`, falling back to the workflow's `index`;
  - reads the workflow definitions once for the `dry_run` label;
  - returns an empty map on any failure.
- **`utils.helper.to_float`** — extracted from six duplicated `isinstance(x, Decimal)` sites across three services (review follow-up).
- **22 tests** covering the window edges, both drop paths, both matching paths, direction parsing, dry-run labelling, multiple and conflicting triggers, and the storage-unavailable path.

## Corrections made during review

- **`order_code` is the asset code, and is now the primary join key.** My first pass treated it as an opaque CFD instrument name and joined only through `Workflow.index`. The repo owner corrected this, and `WorkflowEngine._get_market` backs it up — it market-detects via `workflow.cfd.lower().endswith(":xnys")`, which only works on a `CODE:exchange` value. `index` is kept as a fallback; the workflow read stays because `dry_run` has no other source.
- **Direction parsing uses the enum's own `get_value`.** `workflow_orders` stores `order_direction` as the enum *name* (`WorkflowEngine` writes `.name`), so `Direction("BUY")` raises. `EnumWithGetValue.get_value` compares case-insensitively on the value and resolves both `"BUY"` and `"Buy"`.

## Deviations from the plan, both deliberate

- **One `get_all_workflows()` read instead of per-trigger `get_workflow_by_id`.** The workflows table is one item per workflow, so one scan replaces N lookups and N failure points.
- **The collector takes `alerts` as a third parameter.** The planning docs showed a two-argument signature; that could not satisfy FR-002, since the output is keyed by `Alert.id` and the alert set is the domain of the result. The docs are corrected in this PR.

## Verification

- `poetry run pytest` — 895 passed, 10 skipped (baseline 873; +22 new).
- `poetry run flake8` — clean. `poetry run black . && poetry run isort .` — applied.
- `poetry run mypy .` — 4 errors, all pre-existing `aioboto3` missing-stub notes, unchanged by this PR.

## Not done

Tasks **T002** and **T003** are blocked: this container has no AWS credentials (`UnrecognizedClientException` on any DynamoDB call), and both need live data. They're marked as blocked in `tasks.md`.

T003 is the dump that confirms the recorded codes actually match scanned asset codes. It matters less now that `order_code` is the primary join key — the format-mismatch risk it was insurance against is largely retired — but it's still the cheapest way to confirm the overlap before US1 ships. Unmatched triggers are dropped with every candidate logged, so a mismatch surfaces as log lines rather than silence.

### DIFF
--- a/services/workflow_trigger_service.py
+++ b/services/workflow_trigger_service.py
@@ -29,13 +29,6 @@ def _to_float(value: Any) -> Optional[float]:
     return float(value)
 
 
-def _parse_direction(value: Any) -> Direction:
-    try:
-        return Direction[str(value)]
-    except KeyError:
-        raise SaxoException(f"Unknown order direction: {value!r}")
-
-
 def _asset_keys(alert: Alert) -> List[str]:
     keys = [alert.asset_code.lower()]
     if alert.country_code:
@@ -82,7 +75,7 @@ def _build_trigger(
 
     return WorkflowTrigger(
         workflow_name=order.get("workflow_name", workflow.get("name", "")),
-        direction=_parse_direction(order.get("order_direction")),
+        direction=Direction.get_value(str(order.get("order_direction"))),
         order_price=order_price,
         placed_at=int(order["placed_at"]),
         dry_run=bool(workflow.get("dry_run", False)),
@@ -145,7 +138,7 @@ async def _collect(
 
         try:
             trigger = _build_trigger(order, workflow)
-        except SaxoException as e:
+        except (SaxoException, ValueError) as e:
             logger.warning(f"Dropping malformed trigger: {e}")
             continue
 
--- a/services/workflow_trigger_service.py
+++ b/services/workflow_trigger_service.py
@@ -1,11 +1,11 @@
 import datetime
 import logging
-from decimal import Decimal
 from typing import Any, Dict, List, Optional, Tuple
 from zoneinfo import ZoneInfo
 
 from model import Alert, Direction, WorkflowTrigger
 from utils.exception import SaxoException
+from utils.helper import to_float
 from utils.logger import Logger
 
 PARIS = ZoneInfo("Europe/Paris")
@@ -19,14 +19,6 @@ def _session_window(run_date: str) -> Tuple[int, int]:
         day, datetime.time.min, tzinfo=PARIS
     ).timestamp()
     return int(start), int(datetime.datetime.now(PARIS).timestamp())
-
-
-def _to_float(value: Any) -> Optional[float]:
-    if value is None:
-        return None
-    if isinstance(value, Decimal):
-        return float(value)
-    return float(value)
 
 
 def _asset_keys(alert: Alert) -> List[str]:
@@ -69,7 +61,7 @@ def _build_workflow_map(
 def _build_trigger(
     order: Dict[str, Any], workflow: Dict[str, Any]
 ) -> WorkflowTrigger:
-    order_price = _to_float(order.get("order_price"))
+    order_price = to_float(order.get("order_price"))
     if order_price is None or order_price <= 0:
         raise SaxoException(f"Invalid order_price: {order.get('order_price')}")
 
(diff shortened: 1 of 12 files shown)


services/workflow_trigger_service.py
```diff
@@ -0,0 +1,157 @@
...
+
+def _to_float(value: Any) -> Optional[float]:
+    if value is None:
+        return None
+    if isinstance(value, Decimal):
+        return float(value)
+    return float(value)
+
+
+def _parse_direction(value: Any) -> Direction:
```
RNiveau: The enum already has a method to do that

services/workflow_trigger_service.py
```diff
@@ -0,0 +1,157 @@
...
+
+def _session_window(run_date: str) -> Tuple[int, int]:
+    day = datetime.datetime.strptime(run_date, "%Y-%m-%d").date()
+    start = datetime.datetime.combine(
+        day, datetime.time.min, tzinfo=PARIS
+    ).timestamp()
+    return int(start), int(datetime.datetime.now(PARIS).timestamp())
+
+
+def _to_float(value: Any) -> Optional[float]:
```
RNiveau: Extract a helper here and check everywhere we do this validation

services/workflow_trigger_service.py
RNiveau: Fixed in ec821f3 — now `Direction.get_value(str(order.get("order_direction")))`, and `_parse_direction` is gone.

It's also more tolerant than what I had: `get_value` compares case-insensitively on the *value*, so it resolves both the stored `"BUY"` name form and `"Buy"`, whereas `Direction[value]` only accepted the former.

That invalidated one test — `test_drops_trigger_with_unknown_direction` used `"Buy"` as its invalid input, which now parses correctly. It uses `"HOLD"` instead, with two new tests pinning the tolerance and the `None` case.

---

services/workflow_trigger_service.py
RNiveau: Done in c52f747 — extracted as `utils.helper.to_float`.

Checking everywhere turned up six call sites of the same `isinstance(x, Decimal)` dance across three files, all now on the helper:

- `services/workflow_service.py` — `_convert_all_order_to_item` and `_convert_order_to_list_item`, two conversions each (`order_price`, `order_quantity`). Both also had a function-local `from decimal import Decimal`; those are gone.
- `api/services/alert_digest_service.py` — `ma50_slope`. This one differed from the others: it needed the `None` case, since the attribute is optional. The helper handles both, which is why it returns `Optional[float]`.
- `services/workflow_trigger_service.py` — `order_price` and `trigger_close`.

One behaviour note: the old workflow_service form passed non-`Decimal` values through untouched, so an int stayed an int. `to_float` always widens. The target fields are typed `float` on the Pydantic models, so this is a narrowing of the type surface rather than a change in what the API returns.

895 passed, 10 skipped; flake8 clean; mypy unchanged at the 4 pre-existing `aioboto3` stub notes.

---

PR #697, opened 2026-08-01
feat: teach the triage agent about workflow triggers (spec 028, phase 3 / US1)

RNiveau:
Phase 3 (User Story 1, the MVP) of [spec 028](https://github.com/RNiveau/saxo-order/tree/main/specs/028-triage-workflow-trigger), on top of the collector merged in #696.

Workflow triggers now reach the reasoning path. An asset that fired chart patterns *and* triggered one of your registered workflows today is ranked with that corroboration in view, and the prompt requires the workflow to be named in the rationale so you can see why the asset rose.

## What's here

- **`TriageAgent.synthesize(alerts, triggers=None)`** — takes the resolved trigger map. Stays **synchronous and storage-free**, so the existing tests need no async mocks and the agent never touches DynamoDB.
- **Payload** carries `workflow_triggers` per asset, **only when non-empty** — `{workflow, direction, dry_run, hour}`, with `hour` as Paris-local `HH:MM`. Prices are deliberately omitted: they carry no ranking information and would invite spurious precision.
- **Prompt** teaches the semantics: a trigger is an independent *mechanism* (not another detector output), worth one point of convergence regardless of how many fired; its direction carries authority at least equal to `combo`; a trigger contradicting the pattern read is a red flag; `dry_run` weighs one step lower.
- **`run_alerting`** collects the map inside the existing triage `try/except`.
- **10 new tests** (27 in the file, 905 in the suite).

## `TRIAGE_RESPONSE_SCHEMA` is unchanged, deliberately

Triggers are input-only. They're re-attached from the collector map when building each `TriagedAsset`, never read back from the model response — the same way `patterns` and `ma50_slope` are already re-attached from `grouped` rather than trusted from the reply. A trigger is a fact from storage; echoing it through the model would put a fact at the mercy of generation.

## One rule added beyond the spec

The prompt says a trigger **cannot by itself carry an asset that has nothing else going for it** — it sharpens convergence, it doesn't replace the OPPORTUNITY axis.

This isn't in the spec text, but spec 027 hit exactly this failure with `mm7_break`: a common event that promoted everything it touched, which is why `_TIMING_PATTERNS` exists in the fallback. A workflow trigger has the same shape. The rule is consistent with A-004 (one point, no multiplier) — it just makes the cap explicit where the model can act on it.

## Interaction with spec 027

Main gained `mm7_break` and the timing-pattern distinction while the earlier phases were in flight. The new prompt section is written to sit alongside that work and reuse its vocabulary rather than introduce a parallel scheme — a contradicting trigger is framed as a red flag "exactly like `mm50_touch` on a bearish thesis", and the multiple-triggers-count-once rule is phrased against the `congestion20`/`congestion100` collapse the prompt already teaches.

## Not in this PR

The **deterministic fallback is untouched** — that's US3 (Phase 5), and it means a fallback day currently ignores triggers. Display surfaces (API, Daily Brief, Slack) are US2 (Phase 4).

## Verification

- `poetry run pytest` — 905 passed, 10 skipped (+10 new).
- `poetry run mypy .` — 4 errors, all pre-existing `aioboto3` / `binance.error` missing-stub notes, unchanged.

**T019 is blocked**: end-to-end verification needs AWS credentials this container doesn't have (`UnrecognizedClientException`), same as T002/T003. Its SC-003 half — no asset appears in the brief solely because a workflow fired — is covered by a unit test. The SC-001/SC-002 half depends on live reasoning and can't be asserted offline, so the ranking behaviour itself is unverified against real data.

---


saxo_order/commands/alerting.py
```diff
@@ -550,7 +551,12 @@ async def run_alerting(
                 AnthropicClient(configuration),
                 configuration.triage_slope_threshold,
             )
-            digest = triage_agent.synthesize(all_alerts)
+            triggers = await collect_todays_triggers(
+                dynamodb_client,
+                datetime.datetime.now().strftime("%Y-%m-%d"),
```
RNiveau: Two clocks meet here: this `run_date` is a naive `now()` (UTC inside Lambda), but `_session_window` turns it into a **Paris** midnight and pairs it with `datetime.now(PARIS)` as the end bound.

At the scheduled 18:15 Europe/Paris run the two dates coincide, so nothing is wrong today. But a manual or retried run after 00:00 Paris (23:00 UTC in winter) yields the *previous* UTC date, and the window becomes Paris-midnight-yesterday → now — roughly 25 hours, quietly pulling yesterday's orders into today's digest. `datetime.datetime.now(ZoneInfo("Europe/Paris")).strftime("%Y-%m-%d")` closes that, and matches the timezone the collector already reasons in.

Related but smaller: `synthesize` derives its own `run_date` from a second `now()` call, so the digest's `run_date` and the trigger window's date are computed independently. Passing one value into both would make the pair unambiguous.

---

tests/services/test_alert_triage_service.py
```diff
@@ -372,3 +385,202 @@ def test_format_slack_digest_handles_no_high_assets() -> None:
...
+        assert left.conviction == right.conviction
+        assert left.rank == right.rank
+        assert left.rationale == right.rationale
+        assert left.workflow_triggers == [] == right.workflow_triggers
+
+
+def test_prompt_teaches_the_workflow_trigger_semantics() -> None:
+    assert "workflow_triggers" in TRIAGE_SYSTEM_PROMPT
+    assert "dry_run" in TRIAGE_SYSTEM_PROMPT
+    assert "ONE point of convergence" in TRIAGE_SYSTEM_PROMPT
```
RNiveau: `"ONE point of convergence"` is an assertion on prose. The prompt is the part of this feature most likely to be reworded during tuning, and a rewrite that preserves the rule exactly ("counts once", "a single convergence point") turns this red without any behaviour changing — a test that fails when nothing broke trains you to edit the test rather than read it.

The first two lines are worth keeping: `workflow_triggers` and `dry_run` are the payload keys `_build_payload` actually emits, so those assertions pin a real contract between payload and prompt. I'd drop the third, or replace it with a check that the prompt mentions the key at all.

---

RNiveau (commented):
Reviewed the full diff against `main` plus the collector from #696. The design holds up — nothing here blocks merge. Two inline notes, one latent, one hygiene.

**What's right**

- Keeping `synthesize` synchronous and storage-free is the correct seam. The collector already swallows every failure and returns `{}`, so a DynamoDB outage degrades to exactly the pre-feature digest rather than losing the brief — and `test_synthesize_without_triggers_matches_pre_feature_behaviour` pins that.
- Not extending `TRIAGE_RESPONSE_SCHEMA` is the right call, and the reasoning in the description is the real one: a trigger is a fact from storage, and round-tripping it through generation would make it forgeable by the model. Re-attaching from `grouped` matches how `patterns` and `ma50_slope` are already handled.
- Omitting the key when empty rather than emitting `"workflow_triggers": []` keeps the payload honest — the model sees absence, not an empty container it might reason about.
- Dropping `order_price`/`trigger_close` from the payload is a good call for the stated reason; the prices carry no ranking information and invite the model to anchor on precision it can't use.
- The "cannot by itself carry an asset" rule is beyond the spec but well-argued. The `mm7_break` failure in spec 027 is the same shape — a common event that promotes everything it touches — and the fix belongs in the prompt where the model can act on it, not only in the fallback where `_TIMING_PATTERNS` lives.

**Two things to consider**

1. Mixed timezones between `run_date` and the Paris session window (inline on `alerting.py:556`). Harmless at the current 18:15 Paris schedule, wrong across midnight.
2. A prompt-wording assertion that will break on rewording without any behaviour change (inline on the test).

**One thing to keep in view, not to fix here**

`store_alert_digest` doesn't serialise `workflow_triggers`, so the field lives only in memory for the life of the run — anything reading the digest back from DynamoDB sees nothing. That's T020 in Phase 4, so it's correctly out of scope, but it does mean the trigger currently influences ranking and then vanishes: after this merges, the rationale may name a workflow that the stored digest has no record of. Worth not letting Phase 4 drift far behind.

On the blocked T019 — the honest framing in the description is appreciated. SC-003 is genuinely covered by `test_triggers_never_introduce_an_asset_into_the_digest`; SC-001/SC-002 are not, and can't be offline. Since a trigger's whole job is to move rank, that's the half worth watching on the first live run.

Tests weren't run here — this container's environment can't import `anthropic` — so the 905-passing claim is taken at face value from CI.

---

saxo_order/commands/alerting.py
RNiveau: Good catch — fixed in e5b9ddb, and I took the second half of the suggestion too.

Added `current_run_date()` in `alert_triage_service.py` (Paris-based), and `run_alerting` now computes the date **once** and passes it to both `collect_todays_triggers` and `synthesize(..., run_date=...)`. `synthesize` still defaults to `current_run_date()` when the argument is omitted, so nothing else had to change.

That closes both halves: the window can no longer be built from a UTC date, and the digest's `run_date` and the trigger window's date can no longer be computed from two independent clock reads.

Worth noting the failure was strictly worse than a mislabelled date — `_session_window` would have paired yesterday's Paris midnight with `now()`, so the window silently widened to ~25 hours and pulled the previous session's orders into today's digest. Wrong corroboration, not just a wrong label.

Three tests added: `current_run_date()` matches Paris today, an explicit `run_date` is honoured, and the default path still resolves to it.

---

tests/services/test_alert_triage_service.py
RNiveau: Agreed — dropped in e5b9ddb.

Kept the two key assertions and renamed the test to `test_prompt_documents_the_payload_keys_it_will_receive`, which is what it actually pins: `_build_payload` emits `workflow_triggers` and `dry_run`, and the prompt has to document the keys it will be handed. That's a real payload↔prompt contract and it breaks for a real reason.

The third line was asserting my own wording back at me, which is the failure mode you describe — it would go red on a reword that changed nothing, and the cheapest way out would have been to edit the test.

---

PR #699, opened 2026-08-02
feat: surface workflow triggers in the brief, API and Slack (spec 028, phase 4 / US2)

RNiveau:
Phase 4 (User Story 2) of [spec 028](https://github.com/RNiveau/saxo-order/tree/main/specs/028-triage-workflow-trigger), on top of the reasoning path merged in #697.

**This closes the gap flagged in the #697 review.** Until now a trigger influenced ranking and then vanished — `store_alert_digest` didn't serialise it, so a stored digest could carry a rationale naming a workflow it had no record of. Corroboration now survives the run and reaches every surface the brief is read on.

## What's here

- **Persistence** — `store_alert_digest` serialises `workflow_triggers` per asset, with `direction` as the enum name to match `workflow_orders`. The key is **omitted entirely** when empty, so items for uncorroborated assets are byte-identical to before.
- **API** — `WorkflowTriggerResponse` plus an optional field on `TriagedAssetResponse`, per the contract in `contracts/alert-digests.openapi.yaml`. A **missing** key hydrates to `None`, never an error, so the digests already in DynamoDB read back cleanly (A-005, no backfill). Decimals widen through the existing `to_float`.
- **Daily Brief** — a trigger line per corroboration: direction arrow, workflow name, Paris-local time, and a `dry run` chip when applicable. Renders **nothing at all** when the list is empty — no placeholder, no empty section.
- **Slack** — corroborated names get a ⚡ in the `Top:` line, plus a single count line. No per-trigger listing (FR-018). Noise-tier assets are excluded from the count: a trigger on something the agent filtered out isn't news.

## A bug worth knowing about

Inserting `_serialize_triaged_asset` directly above `store_alert_digest` silently stole its `@_dynamo_operation` decorator — the new synchronous method became a coroutine, and the persistence path lost its error handling and timing instrumentation. Four tests failed immediately with `coroutine 'DynamoDBClient._serialize_triaged_asset' was never awaited`, which is how I found it.

Fixed by restoring the decorator. Flagging it because that decorator sits bare above the method it wraps, so anything inserted at that line inherits it — an easy trap to hit again.

## Verification

- `poetry run pytest` — 918 passed, 10 skipped (+13 new).
- `poetry run mypy .` — 4 errors, all pre-existing missing-stub notes, unchanged.
- `npm run build` — succeeds. `npm run lint` — **0 errors**; the 3 warnings are pre-existing in `AssetDetail.tsx` / `SearchResults.tsx`, none in files touched here.

Note on the frontend gates: `node_modules` wasn't installed in this container, so the first `npm run lint` failed on module resolution rather than on the code. I ran `npm ci` and re-ran both — the results above are from the real runs.

## Not done

**T029 is blocked** — rendering against a real digest needs AWS credentials this container doesn't have. The component compiles and the TypeScript interfaces mirror the Pydantic models field-for-field, but it has **not been seen in a browser**. That's the third blocked verification task (T003, T019, T029), all the same cause.

Phase 5 (US3, the deterministic fallback) remains: a fallback day still ignores triggers entirely.

---

### DIFF
--- a/api/models/alert_digest.py
+++ b/api/models/alert_digest.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional
+from typing import Dict, List, Literal, Optional
 
 from pydantic import BaseModel, Field
 
@@ -7,7 +7,7 @@ class WorkflowTriggerResponse(BaseModel):
     workflow_name: str = Field(
         description="Name of the workflow that fired today"
     )
-    direction: str = Field(
+    direction: Literal["BUY", "SELL"] = Field(
         description="Order direction as the Direction enum name (BUY/SELL)"
     )
     order_price: float = Field(
--- a/api/models/alert_digest.py
+++ b/api/models/alert_digest.py
@@ -3,6 +3,26 @@ from typing import Dict, List, Optional
 from pydantic import BaseModel, Field
 
 
+class WorkflowTriggerResponse(BaseModel):
+    workflow_name: str = Field(
+        description="Name of the workflow that fired today"
+    )
+    direction: str = Field(
+        description="Order direction as the Direction enum name (BUY/SELL)"
+    )
+    order_price: float = Field(
+        description="Price of the order the workflow produced"
+    )
+    trigger_close: Optional[float] = Field(
+        default=None,
+        description="Market close when the workflow fired, if recorded",
+    )
+    placed_at: int = Field(description="Epoch seconds when the workflow fired")
+    dry_run: bool = Field(
+        description="True when the rule fired without committing capital"
+    )
+
+
 class TriagedAssetResponse(BaseModel):
     asset_code: str = Field(description="Asset symbol or ticker")
     asset_description: str = Field(description="Human-readable asset name")
@@ -27,6 +47,13 @@ class TriagedAssetResponse(BaseModel):
     )
     tradingview_url: Optional[str] = Field(
         default=None, description="Custom TradingView URL if configured"
+    )
+    workflow_triggers: Optional[List[WorkflowTriggerResponse]] = Field(
+        default=None,
+        description=(
+            "Same-day workflow firings that corroborated this asset. "
+            "Omitted when the asset has none."
+        ),
     )
 
 
--- a/api/services/alert_digest_service.py
+++ b/api/services/alert_digest_service.py
@@ -6,6 +6,7 @@ from cachetools import TTLCache
 from api.models.alert_digest import (
     AlertDigestResponse,
     TriagedAssetResponse,
+    WorkflowTriggerResponse,
 )
 from client.aws_client import DynamoDBClient
 from utils.helper import to_float
@@ -91,4 +92,23 @@ class AlertDigestService:
             patterns=asset["patterns"],
             ma50_slope=to_float(asset.get("ma50_slope")),
             tradingview_url=tradingview_links.get(asset_id),
+            workflow_triggers=self._to_workflow_triggers(asset),
         )
+
+    def _to_workflow_triggers(
+        self, asset: Dict[str, Any]
+    ) -> Optional[List[WorkflowTriggerResponse]]:
+        triggers = asset.get("workflow_triggers")
+        if not triggers:
+            return None
+        return [
+            WorkflowTriggerResponse(
+                workflow_name=trigger["workflow_name"],
+                direction=trigger["direction"],
+                order_price=to_float(trigger["order_price"]),
+                trigger_close=to_float(trigger.get("trigger_close")),
+                placed_at=int(trigger["placed_at"]),
+                dry_run=bool(trigger["dry_run"]),
+            )
+            for trigger in triggers
+        ]
--- a/client/aws_client.py
+++ b/client/aws_client.py
@@ -12,7 +12,7 @@ import aioboto3
 import boto3
 from botocore.exceptions import ClientError
 
-from model import Alert, AlertDigest, AlertType
+from model import Alert, AlertDigest, AlertType, TriagedAsset
 from utils.json_util import dumps_indicator, hash_indicator
 from utils.logger import Logger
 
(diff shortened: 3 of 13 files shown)


RNiveau:
## Review

Reviewed the full diff, ran the suite and the lint gates locally on `be6a758`, and checked the contract file against the implementation.

The shape of the change is right, and the two things most likely to break existing data are both handled correctly: `store_alert_digest` omits the key entirely when there are no triggers (verified by `test_omits_the_key_when_the_asset_has_no_triggers`), and `_to_workflow_triggers` treats a missing key as `None` rather than a `KeyError`. `contracts/alert-digests.openapi.yaml` declares `workflow_triggers` as `type: [array, "null"]`, so emitting `"workflow_triggers": null` (which is what Pydantic does here, not omission) is inside the contract — worth noting since the prose above that schema says "omitted entirely". The `noise` exclusion also matches between surfaces: Slack skips `Conviction.NOISE`, and `DailyBriefCarousel` already filters `conviction !== 'noise'` before rendering rows, so a trigger on a filtered asset can't leak into the UI either.

### 1. CI is red — and it isn't this PR (blocking anyway)

`backend` fails on `tests/services/test_workflow_trigger_service.py::test_drops_orders_placed_after_now`. I reproduced it on this branch **and on `origin/main`**, so it came in with #697, not here — but it blocks the merge regardless.

It's a wall-clock time bomb, not a logic bug. The test file hardcodes `RUN_DATE = "2026-08-01"` and builds the "future" order as `at_hour(14, 30) + 24 * 3600` → 2026-08-02 14:30 Paris. `_session_window` ends at `datetime.now(PARIS)`, so the test asserted something true only until real time passed that instant. It did, today at 12:30 UTC. The PR body's "918 passed" was accurate when it was run; the same command now gives `1 failed, 917 passed, 10 skipped`.

The narrow fix (derive `RUN_DATE` from today) only moves the bomb — `order()`'s default `at_hour(14, 30)` would then fail whenever the suite runs before 14:30 Paris. The durable fix is to make "now" injectable in `_session_window` (an optional `now` parameter, or a module-level clock the test patches) so the window is deterministic. Happy to open that as a separate PR against `main` if you'd rather not carry it here.

### 2. Slack: a WATCH-tier corroboration is counted but never named

`⚡` is appended only inside the `Top:` line, which is built from `high_assets`. The count line, though, counts every non-noise asset with triggers. Your own test encodes the gap:

```python
_triaged("SAN", Conviction.HIGH, 1, [_trigger()]),
_triaged("AI", Conviction.WATCH, 2, [_trigger("AI rule")]),
...
assert "⚡ 2 workflow-corroborated assets" in message
```

The resulting message shows one ⚡ next to SAN and then claims two corroborated assets, with nothing pointing at the second. A reader can't tell whether they miscounted or whether something is hidden. Two clean ways out, both cheap: count only `Conviction.HIGH` so the number matches the visible marks, or keep the wider count and say so (`⚡ 2 workflow-corroborated (1 in Top)`). FR-018 rules out a per-trigger listing, not naming the asset — so this is a wording call, not a spec conflict.

### 3. `_serialize_triaged_asset` should be module-level

You flagged the decorator trap in the PR body — good catch, and the diagnosis is right. But the fix restores the hazard for the next person. The method never touches `self`, so hoisting it to a module-level `def _serialize_triaged_asset(asset: TriagedAsset) -> Dict[str, Any]:` removes the trap structurally instead of relying on the next author reading the note. That's also how `services/workflow_trigger_service.py` does it (`_build_trigger`, `_session_window`), so it's the house style rather than a new convention.

### 4. `direction` is unconstrained at the API boundary

`WorkflowTriggerResponse.direction` is a bare `str`, while the contract declares `enum: [BUY, SELL]` and the frontend narrows to `'BUY' | 'SELL'`. The consequence is in the JSX:

```tsx
{trigger.direction === 'BUY' ? '▲' : '▼'} {trigger.direction}
```

Anything that isn't exactly `BUY` renders as a **sell** arrow, and `direction.toLowerCase()` yields a class that matches no rule, so it loses its colour too — a wrong direction shown confidently, which is the worst failure mode for this particular field. Nothing writes a bad value today (`_serialize_triaged_asset` uses `.name`, and `record_workflow_order` already stores `.name` via `engines/workflow_engine.py:155`), so this is defence in depth. `Literal["BUY", "SELL"]` on the Pydantic field turns a future mismatch into a loud validation error at the boundary instead of a silent ▼.

### 5. Nit: `key={index}`

`key={index}` on the trigger rows works because the list is static per render, but `key={`${trigger.workflow_name}-${trigger.placed_at}`}` is stable and free.

### On T029

The blocked-verification note is the right call, and stating it plainly beats a checkbox that implies more than was done. Worth saying that `npm run build` passing means the types line up, not that the row renders — the `fr-FR` / `Europe/Paris` formatting in `formatTriggerHour` in particular is the kind of thing that's only really confirmed by looking at it.

---

None of 2–5 are merge blockers on their own; item 1 is, and it's inherited. My suggestion: land the `_session_window` clock fix first (here or separately) to get CI green, fold in 2 and 3, and treat 4 and 5 as optional.

---

RNiveau:
All five addressed. Items 2–5 in b41cbdd; item 1 was already in 7c6ef08, pushed a few minutes before this review landed.

**1. CI — fixed, and we converged on the same fix.** `_session_window` now reads the clock through `_now_paris()`, which the tests pin to 18:15 Paris via an autouse fixture. You're right that deriving `RUN_DATE` from today only moves the bomb — `at_hour(14, 30)` would then fail whenever the suite ran before 14:30 Paris. You're also right that it came in with #697 rather than here.

Worth adding: this is the second time-bomb assertion of mine you've caught, after the prompt-wording one on #697. Same category — a test that goes red when nothing broke.

**2. Slack — took neither option, for a reason.** Counting only `HIGH` would make the number match the marks but silently drop a WATCH corroboration; keeping the count with a parenthetical explains the gap without closing it. Instead the WATCH ones are now named on their own line:

```
Top: Sanofi (SAN) ⚡, Air Liquide (AI)
⚡ Also corroborated: Orange (ORA)
```

Every bolt points at a named asset, nothing is hidden, and it's still not a per-trigger listing. The test now asserts `message.count("⚡") == 2` so a future change can't reintroduce an unresolvable mark.

**3. Hoisted to module level.** Agreed — restoring the decorator fixed the instance but left the trap. It never touched `self`, and it now sits beside `_build_trigger` / `_session_window` in style terms.

**4. `Literal["BUY", "SELL"]` applied.** The failure mode you describe is the convincing part: not a crash but a confident ▼ on a buy, plus a class that matches no CSS rule so it loses its colour too.

**5. Stable key applied** — `${workflow_name}-${placed_at}`.

**One correction to my own PR body.** You're right that Pydantic emits `"workflow_triggers": null` rather than omitting the key. The description said "omitted entirely" for both layers; that's accurate for the DynamoDB item (`_serialize_triaged_asset` only sets the key when non-empty) but not for the API response. The contract's `type: [array, "null"]` covers it and the frontend's `?.` handles it, so the behaviour is right — the prose was wrong.

**On T029** — agreed, and I'd single out `formatTriggerHour`. `toLocaleTimeString('fr-FR', { timeZone: 'Europe/Paris' })` typechecks regardless of what it actually renders, so a passing build says nothing about whether that column reads correctly.

918 passed, 10 skipped; flake8 clean; mypy unchanged at the 4 pre-existing stub notes; `npm run build` succeeds and `npm run lint` reports 0 errors.

---

PR #700, opened 2026-08-02
feat: count workflow triggers in the deterministic fallback (spec 028, phase 5 / US3)

RNiveau:
Phase 5 (User Story 3) of [spec 028](https://github.com/RNiveau/saxo-order/tree/main/specs/028-triage-workflow-trigger), on top of the surfaces merged in #699. **This completes all three user stories.**

A degraded day no longer discards the strongest signal available: when reasoning is unavailable, the deterministic fallback now counts a workflow trigger toward confluence instead of ignoring it.

## What's here

- **`_confluence_points(entry)`** — structural pattern families plus one point if any trigger is attached. **One point however many fired** (A-004): corroboration breaks ties, it doesn't multiply.
- Used by both `_fallback_conviction` and the `_fallback_digest` ordering key, so tier and rank agree.
- **`_fallback_rationale`** names the workflow and its direction, and marks dry runs.
- **7 new tests**.

## The tier rule, and why it matches US1

`1 structural pattern + 1 trigger → HIGH`, but **a trigger alone cannot reach HIGH** — that still needs two structural patterns, or one plus a trigger. An asset whose only pattern is a timing signal (`mm7_break`, excluded from `_structural_families`) plus a trigger tops out at WATCH.

That mirrors the prompt rule added in #697 — "a trigger cannot by itself carry an asset that has nothing else going for it." Had the two paths disagreed, a fallback day would silently reorder the brief for reasons the trader couldn't see from the output.

## A trap worth recording

My first version used the new confluence total in the rationale string, producing `"2 pattern(s): combo"` when only one pattern had fired. Inaccurate on its own — but worse, it would have changed the wording on **no-trigger days too**, breaking SC-004 ("on days with no same-day triggers, the brief is indistinguishable from the pre-feature brief").

The rationale therefore keeps the **structural pattern count** and appends the workflow as its own clause. `test_fallback_rationale_unchanged_when_nothing_is_corroborated` pins the exact pre-feature string.

## A test I deleted

`test_fallback_ignores_triggers_for_now` asserted that the fallback ignored triggers — the Phase 3 state, and precisely what this phase changes. I replaced it with `test_fallback_carries_triggers_onto_the_triaged_asset`, which keeps the half that still holds.

Flagging it explicitly: "the change made a test fail so I removed it" deserves scrutiny. Here the test was scaffolding with an expiry date and its name said so, but it's worth checking rather than taking on trust.

## Verification

- `poetry run pytest` — 925 passed, 10 skipped (+7 new).
- `poetry run mypy .` — 4 errors, all pre-existing missing-stub notes, unchanged.

## Still unverified against live data

All three verification tasks remain blocked on AWS credentials this container doesn't have: **T003** (do recorded codes actually match scanned assets), **T019** (does a corroborated asset really rank first), **T029** (does the trigger row render correctly).

The feature is complete and internally consistent, but no part of it has run against a real workflow order or been seen in a browser. Worth doing before relying on the first live brief.

---

### DIFF
--- a/services/alert_triage_service.py
+++ b/services/alert_triage_service.py
@@ -281,7 +281,7 @@ class TriageAgent:
         ordered = sorted(
             grouped.values(),
             key=lambda entry: (
-                len(self._structural_families(entry["patterns"])),
+                self._confluence_points(entry),
                 self._abs_slope(entry["ma50_slope"]),
             ),
             reverse=True,
@@ -317,12 +317,25 @@ class TriageAgent:
             fallback_used=True,
         )
 
+    def _confluence_points(self, entry: Dict[str, Any]) -> int:
+        """Independent points of evidence backing this asset.
+
+        A workflow trigger is a mechanism separate from the detectors, so it
+        adds one point - and only one, however many fired (A-004). It cannot
+        reach HIGH on its own: that still needs two structural patterns, or
+        one plus a trigger.
+        """
+        points = len(self._structural_families(entry["patterns"]))
+        if entry["workflow_triggers"]:
+            points += 1
+        return points
+
     def _fallback_conviction(self, entry: Dict[str, Any]) -> Conviction:
-        structural_count = len(self._structural_families(entry["patterns"]))
-        if structural_count >= 2:
+        points = self._confluence_points(entry)
+        if points >= 2:
             return Conviction.HIGH
         if (
-            structural_count >= 1
+            points >= 1
             and self._abs_slope(entry["ma50_slope"]) >= self.slope_threshold
         ):
             return Conviction.WATCH
@@ -332,8 +345,22 @@ class TriageAgent:
         patterns = ", ".join(p.value for p in entry["patterns"])
         slope = entry["ma50_slope"]
         slope_text = f", slope {slope:.1f}%" if slope is not None else ""
+        # Deliberately the structural pattern count, not _confluence_points:
+        # on a day with no triggers this string must be byte-identical to the
+        # pre-feature fallback (SC-004). The workflow is named in its own
+        # clause instead.
         structural_count = len(self._structural_families(entry["patterns"]))
-        return f"{structural_count} pattern(s): {patterns}{slope_text}"
+        trigger_text = ""
+        if entry["workflow_triggers"]:
+            trigger_text = "; workflow " + ", ".join(
+                f"{trigger.workflow_name} ({trigger.direction.value}"
+                f"{', dry run' if trigger.dry_run else ''})"
+                for trigger in entry["workflow_triggers"]
+            )
+        return (
+            f"{structural_count} pattern(s): "
+            f"{patterns}{slope_text}{trigger_text}"
+        )
 
     def _pattern_families(self, patterns: List[AlertType]) -> set[AlertType]:
         return {_PATTERN_FAMILY.get(p, p) for p in patterns}
--- a/services/alert_triage_service.py
+++ b/services/alert_triage_service.py
@@ -10,6 +10,7 @@ from model import (
     AlertDigest,
     AlertType,
     Conviction,
+    Direction,
     TriagedAsset,
     WorkflowTrigger,
 )
@@ -317,18 +318,46 @@ class TriageAgent:
             fallback_used=True,
         )
 
+    def _trigger_point(self, entry: Dict[str, Any]) -> int:
+        """Whether the day's workflow triggers earn a confluence point.
+
+        The reasoning path treats a trigger that contradicts the read as a red
+        flag rather than as confluence. This mirrors that: the point is earned
+        only by live triggers that agree with each other and with the
+        direction of the trend they ride. It is never negative - the fallback
+        withholds credit, it does not punish - and never more than one,
+        however many fired (A-004).
+        """
+        live = [t for t in entry["workflow_triggers"] if not t.dry_run]
+        if not live:
+            # dry_run committed no capital: one step below a live trigger,
+            # which in an integer tally means no point at all.
+            return 0
+
+        directions = {trigger.direction for trigger in live}
+        if len(directions) > 1:
+            return 0
+
+        slope = entry["ma50_slope"]
+        if not slope:
+            return 0
+
+        direction = directions.pop()
+        agrees = (direction == Direction.BUY and slope > 0) or (
+            direction == Direction.SELL and slope < 0
+        )
+        return 1 if agrees else 0
+
     def _confluence_points(self, entry: Dict[str, Any]) -> int:
         """Independent points of evidence backing this asset.
 
-        A workflow trigger is a mechanism separate from the detectors, so it
-        adds one point - and only one, however many fired (A-004). It cannot
-        reach HIGH on its own: that still needs two structural patterns, or
-        one plus a trigger.
+        A workflow trigger is a mechanism separate from the detectors, so an
+        agreeing one adds a point. It cannot reach HIGH on its own: that still
+        needs two structural patterns, or one plus a trigger.
         """
-        points = len(self._structural_families(entry["patterns"]))
-        if entry["workflow_triggers"]:
-            points += 1
-        return points
+        return len(
+            self._structural_families(entry["patterns"])
+        ) + self._trigger_point(entry)
 
     def _fallback_conviction(self, entry: Dict[str, Any]) -> Conviction:
         points = self._confluence_points(entry)
@@ -352,11 +381,20 @@ class TriageAgent:
         structural_count = len(self._structural_families(entry["patterns"]))
         trigger_text = ""
         if entry["workflow_triggers"]:
+            # .name, not .value: Slack and the Daily Brief both render the
+            # stored BUY/SELL form, and a fallback rationale sits beside them.
             trigger_text = "; workflow " + ", ".join(
-                f"{trigger.workflow_name} ({trigger.direction.value}"
+                f"{trigger.workflow_name} ({trigger.direction.name}"
                 f"{', dry run' if trigger.dry_run else ''})"
                 for trigger in entry["workflow_triggers"]
             )
+        # "0 pattern(s): mm7_break" would state a count of zero and then list
+        # one. A zero structural count means every pattern was a timing
+        # signal, so say that instead. Only reachable with a trigger attached
+        # (nothing else promotes a timing-only asset), so the no-trigger
+        # string stays byte-identical (SC-004).
+        if structural_count == 0:
+            return f"timing only: {patterns}{slope_text}{trigger_text}"
         return (
             f"{structural_count} pattern(s): "
             f"{patterns}{slope_text}{trigger_text}"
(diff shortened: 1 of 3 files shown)


RNiveau:
## Review

Reviewed `3671f49`, ran the suite and the gates, and exercised the fallback directly across the edge scenarios rather than only reading it. CI is green and I reproduce it: **925 passed, 10 skipped**, flake8/black/isort clean, mypy 4 pre-existing missing-stub notes.

Two things I want to confirm before the findings, because both were flagged in the PR body and both check out:

**SC-004 holds.** A no-trigger fallback still produces `'1 pattern(s): combo, slope 3.0%'` — byte-identical. Keeping `structural_count` out of `_confluence_points` in the rationale was the right call, and the comment explaining why is worth its space.

**The deleted test is fully compensated.** `test_fallback_ignores_triggers_for_now` asserted three things; the replacement keeps two, and the third (the conviction) is now asserted with the *identical* fixture setup — same `_alert("SAN", AlertType.COMBO, 3.0)`, same `{"SAN_xpar": [_trigger()]}` — in `test_fallback_counts_a_trigger_as_one_confluence_point`, just with the new expected value. Nothing was dropped, only re-pointed. Flagging it for scrutiny was right; it survives the scrutiny.

I also confirmed the HIGH bar you describe: there is no path to HIGH without at least one structural pattern, because `points >= 2` needs `structural >= 1`.

### 1. A trigger that *contradicts* the read still promotes it

This is the finding I'd most want you to look at. `_confluence_points` adds its point unconditionally, with no comparison against direction or slope sign. Actual output from the fallback:

```
combo, slope +3.0, BUY trigger                    high   1 pattern(s): combo, slope 3.0%; workflow SAN breakout H1 (Buy)
combo, slope +3.0, SELL trigger (contradicts)     high   1 pattern(s): combo, slope 3.0%; workflow SAN breakout H1 (Sell)
```

Bullish combo on a +3% MA50 slope, and the trader's own rule fired **Sell** — the fallback treats that exactly like agreement and lifts it to HIGH. Worse, it wins the ranking:

```
rank 1  SAN  high   1 pattern(s): combo, slope 3.0%; workflow SAN breakout H1 (Sell)
rank 2  AI   watch  1 pattern(s): combo, slope 3.0%
```

The asset whose evidence *conflicts* outranks the one whose evidence is clean. Your own prompt is explicit that this is backwards:

> A trigger pointing AGAINST the pattern-and-trend read is a RED FLAG on that read […] not "confluence that overrides a conflict". The trader's own rule disagreeing with your reading is evidence your reading is wrong, not evidence to discount.

The PR body argues the tier rule matches US1 so that "a fallback day [doesn't] silently reorder the brief for reasons the trader couldn't see." That reasoning is sound — it just doesn't extend as far as claimed. On the HIGH bar the two paths agree; on direction they invert. And the rationale string names the direction, so a reader can *see* `(Sell)` sitting under a HIGH tier on a bullish read — visible, but as a contradiction the system apparently endorsed.

To be fair to the diff: FR-013 says only "count an attached trigger as one additional independent point", so this is spec-literal. The gap is between the fallback and the prompt's nuance, and the PR invites that comparison itself.

### 2. `dry_run` gets a full point

Same shape, second axis. The prompt says weigh a dry-run trigger "ONE STEP LOWER than a live trigger"; the fallback gives it the identical point:

```
combo, slope +3.0, DRY-RUN trigger    high    1 pattern(s): combo, slope 3.0%; workflow SAN breakout H1 (Buy, dry run)
```

A rule that committed no capital lifts an asset to the top tier on a degraded day. `test_fallback_rationale_marks_a_dry_run_workflow` pins the string but not the tier, so this isn't currently a decision the tests record either way.

For 1 and 2 together, three options, in increasing effort: leave the behaviour and **say so in the `_confluence_points` docstring** (it currently defends the HIGH bar but is silent on direction and dry-run, which reads as symmetry that isn't there); or gate the point on a live trigger; or gate it on direction agreement with the slope sign. The docstring alone would satisfy me — silent divergence is the part worth fixing, and getting it wrong in the direction of "flags a conflict" is cheap on a fallback day.

### 3. `"0 pattern(s)"` reaches the brief

The WATCH gate changed shape: it was `structural_count >= 1`, it's now `points >= 1`, which a trigger satisfies with **zero** structural patterns. So a timing-only asset flips NOISE → WATCH on the trigger alone:

```
mm7_break only, slope +3.0, no trigger     noise   ''
mm7_break only, slope +3.0, BUY trigger    watch   '0 pattern(s): mm7_break, slope 3.0%; workflow SAN breakout H1 (Buy)'
```

`test_fallback_counts_several_triggers_as_one_point` asserts this WATCH outcome, so the promotion looks deliberate — but the string a trader reads is `0 pattern(s): mm7_break`, which states a count of zero and then lists one. It reads like a bug even though it's the structural/timing distinction leaking into user-facing text. Either render it from `len(entry["patterns"])` (which would break the SC-004 byte-identity you deliberately protected — so probably not), or special-case the zero-structural wording, e.g. `"timing only: mm7_break, slope 3.0%; workflow …"`. Worth checking the flat-trend case is intended too: `mm7_break + trigger + slope 0.1` correctly stays NOISE, so the opportunity axis still gates it.

### 4. Nit: `Buy` here, `BUY` everywhere else

`_fallback_rationale` uses `trigger.direction.value` → `(Buy)`, matching the LLM payload. But #699 stores and serves `.name` → `BUY`, which is what Slack and the Daily Brief row show. On a fallback day the same field appears in two casings on the same screen. Harmless, one character to align if you want it consistent.

---

Nothing here is a correctness bug in the sense of crashing or corrupting data, and CI is green. #3 is cosmetic-but-visible, #4 is a nit. **#1 is the one I'd act on before this runs on a live degraded day** — even just as a docstring note recording that direction and dry-run are deliberately not weighed.

Still worth repeating what the PR body already says plainly: T003, T019 and T029 remain unverified against live data, so the confluence rule has never been exercised against a real recorded workflow order.

---

RNiveau:
All four fixed in c237cb2. I went past the docstring on #1 and #2 — you'd have accepted documenting it, but my own PR body claimed the two paths agree on ranking, and that claim was false. Recording a known inversion is weaker than removing one.

First, reproduced exactly as you described:

```
BUY (agrees)           -> high   1 pattern(s): combo, slope 3.0%; workflow R (Buy)
SELL (contradicts)     -> high   1 pattern(s): combo, slope 3.0%; workflow R (Sell)
DRY-RUN                -> high   1 pattern(s): combo, slope 3.0%; workflow R (Buy, dry run)
```

Now:

```
BUY agrees (+slope)          -> high   1 pattern(s): combo, slope 3.0%; workflow R (BUY)
SELL contradicts (+slope)    -> watch  1 pattern(s): combo, slope 3.0%; workflow R (SELL)
SELL agrees (-slope)         -> high   1 pattern(s): combo, slope -3.0%; workflow R (SELL)
DRY-RUN                      -> watch  1 pattern(s): combo, slope 3.0%; workflow R (BUY, dry run)
conflicting pair             -> watch  1 pattern(s): combo, slope 3.0%; workflow R (BUY), S (SELL)
timing only + trigger        -> watch  timing only: mm7_break, slope 3.0%; workflow R (BUY)
```

**1. Direction now gates the point.** `_trigger_point` awards it only to live triggers that agree with each other *and* with the sign of `ma50_slope`. Slope is the only directional read the fallback has, so it's the only thing a trigger can be checked against.

One deliberate limit: a contradiction **withholds** the point, it doesn't subtract one. The prompt calls it a red flag on the read, which in the reasoning path means reconsidering the thesis — the fallback has no thesis to reconsider, so demoting below the no-trigger baseline would be inventing a judgement the tier system can't express. It now ranks level with the same asset carrying no trigger, which is the honest position.

Conflicting triggers on one asset also earn nothing, matching the prompt's "internally inconsistent signal, not doubled conviction."

**2. `dry_run` earns nothing.** "One step lower than a live trigger" has no fractional representation in an integer tally, and the tier outcome is exactly what one step lower should mean: `1 structural + dry-run` stays WATCH instead of reaching HIGH. It still appears in the rationale, so it's visible without being credited.

**3. `timing only:` replaces `0 pattern(s):`.** You're right that it read like a bug. I didn't switch to `len(entry["patterns"])` for the reason you anticipated — it would break the SC-004 byte-identity. Worth noting the branch is only reachable *with* a trigger attached, since nothing else promotes a timing-only asset out of NOISE, so the no-trigger string is untouched by construction.

**4. Casing aligned** to `.name` → `(BUY)`. The LLM payload keeps `.value`, since the prompt documents the field as `"Buy"/"Sell"` — different audience, different contract.

Eight new tests, including the one that matters most: `test_fallback_contradicting_trigger_does_not_outrank_a_clean_read`.

933 passed, 10 skipped; flake8 clean; mypy unchanged at the 4 pre-existing stub notes.

Thanks for exercising the fallback across scenarios rather than reading it — #1 was invisible in the diff and obvious in the output table. And your point about T003/T019/T029 still stands: the confluence rule, including this fix, has never run against a real recorded workflow order.

---

RNiveau:
## Re-review of `c237cb2`

All four points addressed, and the two substantive ones took the stronger option rather than the docstring note I'd have settled for. Re-ran the same scenarios against the new head:

| Scenario | Before | Now |
|---|---|---|
| combo +3.0, BUY trigger (agrees) | high | **high** |
| combo +3.0, SELL trigger (contradicts) | high | **watch** |
| combo −3.0, SELL trigger (agrees) | — | **high** |
| combo +3.0, dry-run trigger | high | **watch** |
| mm7_break only +3.0, BUY trigger | `0 pattern(s): mm7_break…` | `timing only: mm7_break…` |
| combo +3.0, no trigger | `1 pattern(s): combo, slope 3.0%` | unchanged |

A contradicting trigger now lands the asset exactly where it would have been with no trigger at all — credit withheld, not punished, which is the right reading of "red flag" for a deterministic tally. The `_trigger_point` docstring says that explicitly, so the next reader won't have to infer it.

`933 passed, 10 skipped` (+8), flake8/black/isort clean, mypy unchanged at the 4 pre-existing stub notes.

Two things I checked that turned out to be non-issues, recorded so nobody re-opens them:

- **`if not slope`** catches `0.0` as well as `None`. Behaviourally identical here — a flat slope agrees with neither direction, so the point would be withheld either way, and `abs(0.0)` fails the WATCH gate regardless. No change needed.
- **Tie ordering.** A contradicting-trigger asset and a clean one now hold equal points and equal slope, so their relative rank falls back to insertion order. That's pre-existing tie behaviour, not something this diff introduced, and the new test correctly asserts on conviction rather than rank.

One judgment call left as-is, flagged rather than requested: `live = [t for t in … if not t.dry_run]` filters dry runs *before* the direction-conflict check, so a live BUY alongside a dry-run SELL is credited as clean agreement — the paper trade's dissent is discarded rather than treated as conflict. Verified: `combo +3.0, live BUY + dry-run SELL → high`. Defensible, since a dry run is meant to weigh lower than a live trigger, and arguably an internally-conflicting pair should still lose the point. Your call; it's a rare shape.

No further findings. The remaining caveat is unchanged and not about this diff: T003, T019 and T029 have still never run against a real workflow order.

---

PR #702, opened 2026-08-03
feat: MM50 direction filter GER40 backtests (G9HICMH / G9HICMD)

RNiveau:
Spec 025, Addendum 5. Two new backtests built on `G9HIC` — the **single-take-profit** impulsive variant, not the two-lot `G9HICD` — with one rule added: **the day trades in at most one direction**.

At 10:00, the 9:00–10:00 reference candle's close is compared to an MA50. Above it the day is long-only, below it short-only, exactly on it untradeable. The two definitions differ **only** in the timeframe that MA50 is read on, so any difference between their results is attributable to the timeframe and nothing else:

| Code | MA50 read on |
|---|---|
| `G9HICMH` | H1 candles (9:00–17:30 cash session) |
| `G9HICMD` | Daily candles (50 closes strictly before the day) |

`G9HIC` is unchanged and remains the unfiltered control run.

## Design notes

- **The filter only refuses an *entry* on the wrong side.** It never re-prices, delays or re-times an entry it allows, and never touches an open position. Both `DirectionSearch` objects keep being fed on every candle, and a refused entry does not trigger the search reset an actual entry does — so the allowed side sees exactly the candle sequence it sees under `G9HIC`. A regression test pins this by interleaving a refused long with a pending short candidate.
- **A day whose MA50 cannot be computed is a `NO_TRADE`**, not a fallback to unfiltered behavior — otherwise a run would silently mix filtered and unfiltered days and the comparison against `G9HIC` would mean nothing. A failed fetch degrades the same way: visibly nothing.
- **Lookahead-safe.** The H1 MA50 ends *with* the reference candle, which has closed by the 10:00 decision; nothing after 10:00 is read. The daily MA50 stops strictly before the trading day.
- **The H1 series is built on the cash session**, not the 9:00–22:00 CFD session these variants trade — the same comparability argument the existing `mm50_slope` / `adx14` columns make. A CFD-session series would put ~13 bars in a day instead of 9, so a "50-period MA" would span four sessions instead of six.
- **Fetched once per range run**, not once per day; the daily variant reuses the series the run already fetches for the regime columns, so it adds no fetch at all. Neither is fetched for a day the cheaper minimum-H1-range filter already rejects.

## Trade-off worth flagging

The comparison price is the **reference candle's close**, not the entry price (a per-day regime read, as specified). An entry can therefore open on the unfavorable side of the MA50 if price crossed back between 10:00 and the breakout. Intended, and recorded in the spec's edge cases.

Also: a daily MA50 barely moves between adjacent days, so `G9HICMD`'s results read as runs of same-direction days rather than as an independent per-day draw.

## Testing

- New `tests/api/services/backtest/test_direction_filter.py` — 33 tests covering the direction decision, the both-searches-still-fed invariant, the unavailable-MA50 path, the lookahead boundaries on both timeframes, fetch shape/ordering, and the definition guard.
- Golden snapshot regenerated: the two new definitions are **added** and **no existing definition's rows move** (verified key-by-key against the previous snapshot). The synthetic golden market gained an H1 branch so the filter runs against real H1 bars there.
- Full suite: 969 passed, 10 skipped. `black` / `isort` / `flake8` clean; `mypy` reports only the 4 pre-existing missing-stub errors in untouched files.

No frontend change — the Backtest menu is registry-driven.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HPNksF5PEVKeMuFtYAxd9V)_


RNiveau:
## Review

Checked out the branch and ran everything locally. The claims in the description hold:

- `poetry run pytest` → **969 passed, 10 skipped**
- `black --check` / `isort --check-only` / `flake8` clean; `mypy .` → only the 4 pre-existing `import-untyped` errors (`binance.error`, `aioboto3` ×3)
- Golden snapshot diffed key-by-key against `4f9cbda`: **added** `{G9HICMH, G9HICMD}`, **changed** `[]`. The "no existing definition's rows move" claim is accurate.

### What I verified in the logic

- **Lookahead boundary (H1).** `_build_h1_from_30m` sets `last.date = first.date`, i.e. an H1 candle carries its *open* time, so `candle.date <= reference_start` in `_h1_candles_through_reference` includes the 9:00–10:00 bar and excludes the 10:00 one. Correct, and matches `get_candles_in_window`'s `start <= date < end` convention for the reference candle itself.
- **MA orientation.** `mobile_average` averages `candles[:period]`, newest-first; both helpers `sort(..., reverse=True)` before returning. Consistent.
- **"Both searches still fed."** Real, not just asserted by the test: feeding happens above the `for side in (LONG, SHORT)` loop, and `self._reset(searches)` runs only when a position actually opened — so a refused side is a pure no-op on search state.
- **Daily series reuse.** `_fetch_daily_candles` anchors at midnight Paris on `end_date`, so `last_session_close` resolves to `end_date - 1`'s close and the run's last day is *not* in the series — harmless, since `_daily_candles_before` wants strictly-prior days anyway. The lead-in (`(end-start).days + 65`) over-counts because calendar days ≥ trading days, so the first day of a long range still has its 50.
- **Single-day path** (`evaluate_day` → self-fetch, min-range short-circuit) is covered by `run_mm50` / `TestFilterOrdering`, not just the range path.

### Findings

**1. `series_unit_time()` is dead code.** `api/services/backtest/direction_filter.py:36` — defined, exported nowhere, called nowhere (including tests); `session_range.py` reads `definition.ma50_direction_filter` directly. Drop it.

**2. `H1_CANDLES_PER_SESSION = 9` duplicates a value `build_candles` derives.** `direction_filter.py:28` hardcodes what `candles_service.build_candles` computes as `close_hour - open_hour + (1 if open_minutes == 0 else 0)` for `EUMarket`. The two are equal today, but if the cash session ever moves they drift apart silently and in the worst direction: an under-sized count means a short series, which means `NO_TRADE` days rather than an error. Worth deriving it from `EUMarket()` (or asserting the equality once).

**3. Silent degradation is the thing I'd most want a signal on.** A failed H1 fetch → `[]` → *every* day of the run reports `NO_TRADE`, and a truncated 30m history at the far end of a long range → the first N days report `NO_TRADE`. In the results table those are indistinguishable from days that were genuinely untradeable — which is exactly the comparison against `G9HIC` this pair exists to support. "Visibly nothing rather than silently unfiltered" is the right call, but as built it's only visible in a log line. A summary-level flag (days scored / days without an MA50) would make the failure legible without a new `DayStatus`.

   Related sizing note: for a 1-year `G9HICMH` range the H1 fetch is `365*9 + 68 ≈ 3.3k` H1 candles → ~360 calendar days → ~17k 30m bars → ~15 paginated Saxo calls. It's once per run and `get_historical_data` paginates fine, but it's not free, and it's the path most likely to hit Saxo's intraday history depth.

**4. Minor: the compared close and the MA come from different aggregation paths.** `h1_candle.close` comes from `CandleSource` (CFD session, `get_candles_in_window`), the MA50 from `build_candles` over 30m bars on the cash session. The reference bar is in both, but built two different ways, so they can differ by a tick. Doesn't matter for the strategy; it does mean the documented "closed exactly on the MA50 → untradeable" branch is effectively unreachable in production. Fine, just worth knowing the third case is theory.

**5. Nothing forbids `combo_entry` + `ma50_direction_filter` at validation time.** Moot today — `StrategySelector.for_definition` raises for any `combo_entry` definition — so only flagging it in case combo lands later and the filter quietly doesn't apply there.

### Verdict

Approach looks sound, and the isolation argument (the filter refuses entries only, never re-times or re-prices) is genuinely enforced by the code, not just by the tests. Nothing here blocks merge — (1) is a trivial delete, (2) and (3) are the two I'd act on. The branch is `behind` main, so it'll want a base merge first.

---

RNiveau:
Thanks — acted on 1, 2 and 5; base merged. Pushed as `b385bbf`.

**1. `series_unit_time()`** — deleted. It was a leftover from an earlier shape where the engine asked the module for the timeframe instead of reading the field.

**2. `H1_CANDLES_PER_SESSION`** — now derived from `EUMarket` with the same expression `build_candles` uses (`close_hour - open_hour + (1 if open_minutes == 0 else 0)`), with the drift argument recorded in the comment. Same value today, so the golden snapshot is byte-identical. The module-level `_CASH_MARKET` also now feeds `paris_reference_window_utc`, so the count and the decision boundary come from one market object rather than two constructions.

**5. `combo_entry` + `ma50_direction_filter`** — added to the `session_range_only` dict in `__post_init__`, so it is rejected at registration with the same "describes the 9h reference range" message as the other eight flags, and added to the parametrized guard test. It genuinely belongs there: the filter compares against the 9h reference candle's *close*, which a combo definition has no notion of.

**4** — agreed, no action. The reference bar reaching the comparison via `CandleSource` and the MA via `build_candles` does make the exactly-on-the-MA50 branch unreachable in practice. It exists so the ordering is deterministic rather than accidental, and the spec's clarification already frames it that way.

**3 — the one I'm not acting on unilaterally.** You're right that "every day is `NO_TRADE`" and "this instrument had a quiet month" are indistinguishable in the results table, and that this is precisely the comparison the pair exists to support. But a days-scored / days-without-MA50 counter on the summary flows into `BacktestRunSummary`, the API response model and the frontend table — user-visible surface beyond what Addendum 5 specifies, and it would show on every backtest, not just these two. Your call on scope; I'd rather ask than widen the spec on my own. Options as I see them:

- a summary-level counter, as you suggested (needs the model + API + frontend row);
- a backend-only signal — one `WARNING` per run naming the count, no schema change;
- leave it, and treat an all-`NO_TRADE` result as the tell.

Verified after the fixes and the base merge: **970 passed, 10 skipped**; `black`/`isort`/`flake8` clean; `mypy` unchanged at the 4 pre-existing `import-untyped` errors; golden snapshot regenerates identically (`G9HICMH` 23 trades / 1067.87, `G9HICMD` 34 / 1160.54).

The sizing note is fair and I've left the fetch as is — once per run, and the alternative (per-day fetches) is strictly worse. If it does hit Saxo's intraday depth on a long range, the failure mode is the one in (3), which is another argument for making it legible.

---

